PR #7, opened 2015-09-16
Improve BCF handling

gusevfe:
Add some functions related to BCF/VCF:

* convert FORMAT tags to string
* convert reference ID to string
* update INFO tags


christopher-schroeder:
Thank you! I published a new version including your changes a few seconds ago.

PR #21, opened 2016-09-19
Use properly typed parameters for constructing readers and writers.

johanneskoester:


johanneskoester:
@pmarks I have renamed your `text` method in `HeaderView` to `as_bytes` in order to improve consistency with other structs and the Rust standard library.

PR #22, opened 2016-11-02
access to raw SAM format header text

pmarks:


pmarks:
@johanneskoester @christopher-schroeder - can you guys take a look?

src/bam/mod.rs
```diff
@@ -507,9 +506,9 @@ impl HeaderView {
     }
 
     /// Retrieve the textual SAM header as a String
-    pub fn text(&self) -> Result<String, FromUtf8Error> {
+    pub fn text(&self) -> Vec<u8> {
```
johanneskoester: What about just returning the byte slice here? I don't think we need the reallocation, do we?

pmarks:
@johanneskoester - yes, good call

johanneskoester:
Great, thank you! Sorry for all the nitpicking :-).

PR #23, opened 2016-11-17
add del_aux wrapper for deleting aux tags

pmarks:
thanks to Wei at 10x.

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -306,15 +306,12 @@ impl Record {
         }
     }
 
-    // Delete auxiliary tag. Return false if tag doesn't exist
-    pub fn del_aux(&self, tag: &[u8]) -> bool {
+    // Delete auxiliary tag.
+    pub fn remove_aux(&self, tag: &[u8]) {
         let aux = unsafe { htslib::bam_aux_get(self.inner, ffi::CString::new(tag).unwrap().as_ptr() as *mut i8 ) };
         unsafe {
-            if aux.is_null() {
-                false
-            } else {
+            if !aux.is_null() {
                 htslib::bam_aux_del(self.inner, aux);
-                true
             }
         }
     }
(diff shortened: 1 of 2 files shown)


pmarks:
@johanneskoester - can you take a look?

src/bam/record.rs
```diff
@@ -306,6 +306,19 @@ impl Record {
         }
     }
 
+    // Delete auxiliary tag. Return false if tag doesn't exist
+    pub fn del_aux(&self, tag: &[u8]) -> bool {
```
johanneskoester: I think it should be `remove_aux` in order to match the signature of HashMap. I also wonder about the return type. It would be more Rusty to return either Option or Result. Option does not make sense here, because we don't return a value. Hence, it should be `Result<(), RemoveAuxError>`, right? `RemoveAuxError` can be defined using the `quick_error` macro as you can see in other places here.

johanneskoester:
Sorry for the late reply. I am travelling at the moment.

src/bam/record.rs
johanneskoester: You are right, it is not an error. Thinking about it, it is more like the [retain method](https://doc.rust-lang.org/beta/std/vec/struct.Vec.html#method.retain). We remove stuff, but no value is returned. Rust does not provide a return type there. Hence, I would vote for doing the same.

johanneskoester: You are right, it is not an error. Thinking about it, it is more like the [retain method](https://doc.rust-lang.org/beta/std/vec/struct.Vec.html#method.retain). We remove stuff, but no value is returned. Rust does not provide a return type there. Hence, I would vote for doing the same.

johanneskoester: Well, forget the issue with the bool return. The bit-set crate does the same. Since that comes from the std actually, I think it is ok to return a bool here. Sorry for the noise.

I still think it should be `remove_aux` though.

pmarks:
@johanneskoester - should be ready to go now.  thanks for the review -- being careful with the design will pay dividends in the future!

johanneskoester:
Thanks a lot!

PR #26, opened 2016-12-09
threaded writing for BAM files

pmarks:
activate htslib support for multi-threaded BAM writing.

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -363,10 +363,10 @@ impl Writer {
     /// # Arguments
     ///
     /// * `n_threads` - number of background writer threads to use
-    pub fn set_threads(&mut self, n_threads: usize) -> Result<(), WriteError> {
+    pub fn set_threads(&mut self, n_threads: usize) -> Result<(), ThreadingError> {
         let r = unsafe { htslib::bgzf_mt(self.f, n_threads as ::libc::c_int, 256) };
         if r != 0 {
-            Err(WriteError::Some)
+            Err(ThreadingError::Some)
         } else {
             Ok(())
         }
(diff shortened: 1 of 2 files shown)


src/bam/mod.rs
```diff
@@ -805,6 +819,56 @@ mod tests {
...
+        println!("{:?}", bampath);
+        {
+            let mut bam = Writer::from_path(
+                &bampath,
+                Header::new().push_record(
+                    HeaderRecord::new(b"SQ").push_tag(b"SN", &"chr1")
+                                            .push_tag(b"LN", &15072423)
+                )
+            ).ok().expect("Error opening file.");
+            bam.set_threads(4).ok();
```
johanneskoester: Can you also add an expect?

src/bam/mod.rs
```diff
@@ -358,6 +358,20 @@ impl Writer {
         Ok(Writer { f: f, header: HeaderView::new(header_record) })
     }
 
+    /// Activate multi-threaded BAM write support in htslib. This should permit faster
+    /// writing of large BAM files.
+    /// # Arguments
+    ///
+    /// * `n_threads` - number of background writer threads to use
+    pub fn set_threads(&mut self, n_threads: usize) -> Result<(), WriteError> {
```
johanneskoester: Maybe have a separate error type for this? ThreadingError::Some? With a description like "error setting threads for multithreaded writing".

src/bam/mod.rs
johanneskoester: If we have the error type I outline above, this can simply be an unwrap call.

pmarks:
@johanneskoester - sorry, forgot about this one for a while.

johanneskoester:
No problem, thank you!

PR #31, opened 2017-02-28
adding SAMWriter

sitag:
I think this does it.Thoughts? Did I miss something?


src/bam/mod.rs
```diff
@@ -403,6 +403,157 @@ impl Drop for Writer {
...
+    }
+
+    /// Read bam file. For each record apply f to it, and write to sam file if f returned Some(true), skip record if Some(false) if None then terminate iteration
+    ///
+    /// # Arguments
+    ///
+    /// * `bamfile` - the bam file to read from
+    /// * `samfile` - the sam file to write
+    /// * `f` - the predicate to apply
+    pub fn from_bam_with_filter<'a, 'b, F>(bamfile:&'a str, samfile:&'b str, f:F) -> Result<(), SAMError> where F:Fn(&record::Record) -> Option<bool> {
```
johanneskoester: I don't think I'd like to have this special case in the codebase. Can you please remove it?

src/bam/mod.rs
```diff
@@ -403,6 +403,157 @@ impl Drop for Writer {
...
+    /// Create a new SAM file at STDOUT.
+    ///
+    /// # Arguments
+    ///
+    /// * `header` - header definition to use
+    pub fn from_stdout(header: &header::Header) -> Result<Self, SAMError> {
+        Self::new(b"-", header)
+    }
+
+    fn new(path: &[u8], header: &header::Header) -> Result<Self, SAMError> {
```
johanneskoester: There is quite some code duplication here. Maybe BAM and SAM can share a common trait?

src/bam/mod.rs
```diff
@@ -403,6 +403,157 @@ impl Drop for Writer {
 }
 
 
+// ---- SAMWriter ------------------
```
johanneskoester: I think this should go into a separate module.

sitag: Sure. Will do that.

src/bam/mod.rs
sitag: Refactoring out the common traits will probably turn into more work than I have time right. Once SAMWriter gets it's own module, it will cleaner.

src/bam/mod.rs
sitag: I thought so too. I was using that function for the SAMWriter test, I will make it private with the there.

src/sam/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+        )
+    };
+    if ret.is_null() {
+        Err(SAMError::IOError)
+    } else {
+        Ok(ret)
+    }
+}
+
+impl SAMWriter {
```
johanneskoester: With the own module, all the `SAM` prefixes should be removed.

src/bin/test_bam_decompress.rs
```diff
@@ -0,0 +1,48 @@
+extern crate rust_htslib;
```
johanneskoester: Can you move this into a proper test case within the new sam module?

sitag: This test is not needed - it's essentially testing that writing sam output to "-" works, i.e. `sam_write1` works (which belong in the underlying library). Also, there isn't an obvious way to redirect output of stdout temporarily over on the `C` side, so it will need to be tested with a subprocess. This is complicated by `cargo test` capturing stdout.  But since this test doesn't belong here anyway, I will remove it.

src/sam/mod.rs
sitag: sure

src/sam/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+use htslib;
+
+use bam::header;
+use bam::record;
+use bam::Reader;
+use bam::Read;
+
+// new on bam::HeaderView is not public
+pub struct HeaderView {
+    inner: *mut htslib::bam_hdr_t,
```
johanneskoester: Can the headerview be taken from bam.rs?

src/sam/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+quick_error! {
+    #[derive(Debug)]
+    pub enum WriterError {
+        IOError {}
+    }
+}
+
+#[test]
```
johanneskoester: Can you put these into a nested test module like suggested in the rust book?

src/sam/mod.rs
sitag: yes, but some methods on it must be made public. this really falls under refactoring out common traits from bam+sam writers. i have made the appropriate methods in bam module public (i suspect that is what is causing the conflicts)

src/sam/mod.rs
sitag: Done, now as you have in `bam/mod.rs`.

src/bam/mod.rs
```diff
@@ -403,6 +403,14 @@ impl Drop for Writer {
...
 
 
+
+
+
+
+
+
+
+
```
johanneskoester: Can you remove these additional lines?

sitag:
That's alright... it would be appreciated more down the road :)

PR #38, opened 2017-05-22
Refactor CIGAR API into views

johanneskoester:
That allow to determine read_pos and end_pos without having to pass a reference to the CIGAR string back to the record object.

This also incorporates code from @dlaehnemann to determine the read position corresponding to a given reference position (prosic/libprosic#6).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: rust
 
+cache: cargo
+
 sudo: false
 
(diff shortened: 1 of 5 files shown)


johanneskoester:
@dlaehnemann you can now use `record.cigar().read_pos(vpos)` to replace your changes in `libprosic::model::sample::prob_read_snv` from PR prosic/libprosic#6.

PR #40, opened 2017-05-29
add support for seek & tell to get virtual offsets ...

pmarks:
from the underlying BGZF handle. 

This ties us to the details of the BGZF struct, because bgzf_tell is unfortunately implemented as a macro in htslib, rather than a method.

One question: pysam uses "fetch" for a position-based query, and seek/tell to set/get the virtual offset of a BAM file.  rust-htslib is using the "seek" for a position-based query. This led to the terms seek_voffset / tell_voffset for the new functionality. Wondering if we should reconsider the naming at all?

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -47,12 +47,13 @@ pub trait Read: Sized {
     /// Return the header.
     fn header(&self) -> &HeaderView;
 
-    // Seek to the give virtual offset in the file
+    // Seek to the given virtual offset in the file
     fn seek(&self, offset: i64) -> Result<(), ReadError> {
         let ret = unsafe { htslib::bgzf_seek(self.bgzf(), offset, libc::SEEK_SET) };
-        match ret { 
-            0 => Ok(()),
-            _ => Err(ReadError::SeekError)
+        if ret == 0 {
+             Ok(())
+        } else {
+            Err(ReadError::SeekError)
         }
     }
 
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -48,7 +48,7 @@ pub trait Read: Sized {
     fn header(&self) -> &HeaderView;
 
     // Seek to the give virtual offset in the file
-    fn seek_voffset(&self, offset: i64) -> Result<(), ReadError> {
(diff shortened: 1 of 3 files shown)


johanneskoester:
Thanks! Out of curiosity: do you have a specific use case?

pmarks:
main use case for us is indexing and seeking sam files sorted by the 10x BX tag -- let us fetch all the reads with a given barcode quickly, for example.

johanneskoester:
I see, that makes a lot of sense. Regarding the naming: you are right, fetch, seek and tell are better names. I am fine with that change.

pmarks:
@johanneskoester  - naming updated

src/bam/mod.rs
```diff
@@ -438,6 +452,12 @@ quick_error! {
         NoMoreRecord {
             description("no more record")
         }
+        FetchError {
+            description("bam fetch failed")
+        }
+        SeekError {
+            description("bam seek failed")
+        }
```
johanneskoester: Why are above two errors needed as part of readerror?

pmarks: not sure what the general principle is, but they are separate kinds of failures.  the fetch can fail if there's a problem with the index, or you pass an unrecongized contig.  the seek can fail if you pass an invalid voffset. Not sure if it's worth having separate errors or not.

johanneskoester: That is of course correct and you define these errors below. My point is that they don't need to be additionally part of readerror. Or are you throwing them anywhere and I just don't see it on my mobile phone?

pmarks: gotcha, i understand now.  fixed.

johanneskoester:
Thanks!

johanneskoester:
@pmarks I had to refactor the implementation into a C-only wrapper for such macros as bgzf_tell. The reason is that the replication of the bitfields in your BGZF struct was not platform independent. Depending on compiler optimizations, it could look different, such that bgzf_tell would give wrong results. Now the test case works fine hopefully.

See commit b49bd7d0.

pmarks:
@johanneskoester  - great, thanks!

PR #43, opened 2017-06-13
Minor suggestions for CIGAR errors & pos parsing.

dlaehnemann:
Before deleting the CigarError and CigarString parsing from my libprosic branch, I went over the implementation here and have some minor suggestions to @johanneskoester's implementation here. Mainly:

1. A distinction between `Unsupported` and `Unexpected` Cigar operations.
2. A mixture of our wordings in the CigarError messages.
3. The question whether to support the `Pad` operation, where the SAMv1 spec is not really clear on how it works and where I am not sure if any software uses it...

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -827,4 +827,192 @@ mod tests {
             println!("{}", op);
         }
     }
+
+    #[test]
+    fn test_cigar_read_pos() {
+        let vpos  = 5; // variant position
+
+        // Ignore leading HardClip
+        // ref:       00 01 02 03 04 05 06 07 08 09 10 11 12 13 14 15
+        // var:                       V
+        // c01: 7H                 M  M
+        // qpos:                  00 01
+        let c01 = CigarStringView {
+            inner: CigarString( vec![Cigar::HardClip(7), Cigar::Match(2)] ),
+            pos: 4
+        };
+        assert_eq!(c01.read_pos(vpos, false, false).unwrap(), Some(1) );
+
+        // Skip leading SoftClip or use as pre-POS matches
+        // ref:       00 01 02 03 04 05 06 07 08 09 10 11 12 13 14 15
+        // var:                       V
+        // c02: 5H2S         M  M  M  M  M  M
+        // qpos:  00        02 03 04 05 06 07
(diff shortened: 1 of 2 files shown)


dlaehnemann:
Just saw from the coveralls-check: proper tests are still missing, something I could model around my tests in libprosic. I'll add these to this pull request tomorrow, so feel free to ignore it until then.

dlaehnemann:
Well, writing and including the tests came up with some corner cases on HardClips where the code did not yet enforce the SAM specification. Also, I found some recent (incoming into the spec master branch) clarifications of how the "Pad" Cigar operation works and implemented this accordingly. Basically, the existing wording that it reflects a "silent deletion from a padded reference" means that it consumes neither read nor reference sequence, which apparently can make sense in cleanly representing multiple alignments with insertions of varying size at the same position...

Also, I added a TODO whether we want to deprecate the `Back` operation in `CigarStringView::end_pos` as well--this would mean changing its return type to pass up the Error. And I changed the wording of the `include_del` parameter of `CigarStringView::read_pos` to reflect the current implementation, but am unsure if this is the intended behaviour.

Please Review away...

src/bam/record.rs
```diff
@@ -606,7 +611,14 @@ impl CigarStringView {
...
                 &Cigar::Equal(l) | &Cigar::Diff(l) => pos += l as i32,
-                &Cigar::Back(l) => pos -= l as i32,
+                &Cigar::Back(l) => pos -= l as i32, // TODO: Don't we want to consistently deprecate this operation? E.g. by returning the following error (this would require a change of this function's signature!)
+                /*
+                {
+                    return Err(CigarError::UnsupportedOperation(
+                        "'back' (B) operation is deprecated according to htslib/bam_plcmd.c and is not in SAMv1 spec".to_owned()
+                    ));
+                }
+                */
```
johanneskoester: Yes, please do that.

johanneskoester:
Very nice! If you would like to deprecate back in end_pos, that would be nice!

dlaehnemann:
Ok, the deprecation of the `Back` Cigar operation is in `CigarStringView::end_pos()`. We have to watch out to make the respective changes in `libprosic` when using the next `rust-htslib` release.

PR #44, opened 2017-07-14
Made CigarStringView more accessible

veldsla:
Because I use the cigarstring in a project I need to be able to create the CigarString and the views. I added a few methods to make this more easy:

 * Added into_view to CigarString
 * Added public constructor to CigarStringView

Updated code and tests.

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -566,6 +566,7 @@ custom_derive! {
 }
 
 impl CigarString {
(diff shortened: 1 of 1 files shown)


johanneskoester:
Thanks! You could also implement `From<CigarString>` for `CigarStringView`. But that can be a separate PR.

veldsla:
I thought about that, but you need to set the position to create the view

johanneskoester:
Oh yes, you are right :-)

PR #46, opened 2017-09-07
Add set_qname

pryvkin10x:
Allows you to set the qname of a bam record. A little bit of refactoring between set() and set_qname().

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -808,6 +808,8 @@ mod tests {
         let mut rec = record::Record::new();
         rec.set_reverse();
         rec.set(names[0], &cigars[0], seqs[0], quals[0]);
+        // note: this segfaults if you push_aux() before set()
+        //       because set() obliterates aux
         rec.push_aux(b"NM", &Aux::Integer(15));
 
         assert_eq!(rec.qname(), names[0]);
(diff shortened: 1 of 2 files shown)


src/bam/record.rs
```diff
@@ -222,6 +218,63 @@ impl Record {
         utils::copy_memory(qual, &mut data[i..]);
     }
 
+    pub fn set_qname(&mut self, new_qname: &[u8]) {
```
johanneskoester: Can you add a doc comment above?

src/bam/mod.rs
```diff
@@ -808,6 +808,8 @@ mod tests {
         let mut rec = record::Record::new();
         rec.set_reverse();
         rec.set(names[0], &cigars[0], seqs[0], quals[0]);
+        // note: this segfaults if you push_aux() before set()
+        //       because set() obliterates aux
```
johanneskoester: Maybe you can also add that to the corresponding docstrings: "push_aux may be never called before set".

src/bam/record.rs
johanneskoester: Will this also invalidate aux?

johanneskoester:
Thanks!

PR #49, opened 2017-12-01
Add static method to parse SAM text into Record objects

pmarks:


src/sam/mod.rs
```diff
@@ -91,6 +75,31 @@ impl Writer {
     }
 }
 
+pub struct SamReader;
+
+impl SamReader {
+    /// Parse a line of SAM data to a Record, given a HeaderView
+    pub fn parse_record(header_view: &HeaderView, sam: &[u8]) -> Result<record::Record, WriteError> {
```
johanneskoester: I think it would make sense to put this as a constructor from_sam into bam::Record.

johanneskoester:
Thank you, this is a very good idea!

PR #50, opened 2017-12-11
icreate Cigar from a bio::alignment::Alignment. 

pmarks:
Not quite ready to merge. We need to release rust-bio 0.15 to include the new alignment changes before this will work.  The rust-bio dependency will then need to be "^0.15".

I've made bio an optional dependency, on by default. Should it be on of off by default? It's a pretty big dependency, so maybe should be off by default?


src/bam/record.rs
```diff
@@ -666,6 +668,64 @@ impl CigarString {
     pub fn into_view(self, pos: i32) -> CigarStringView {
         CigarStringView::new(self, pos)
     }
+
+    /// Calculate the bam cigar from the alignment struct. x is the target string
+    /// and y is the reference.
+    #[cfg(feature = "bio")]
+    pub fn from_alignment(alignment: &Alignment, hard_clip: bool) -> Self {
```
johanneskoester: Maybe implement the `From` trait instead (see [here](https://doc.rust-lang.org/std/convert/trait.From.html))?

src/bam/record.rs
```diff
@@ -1078,3 +1137,94 @@ mod tests {
         assert_eq!(rec.pos(), clone.pos());
     }
 }
+
+#[cfg(feature = "foo")]
```
johanneskoester: I guess this should be `feature = "bio"`?

johanneskoester:
I will release a new version of rust-bio today. Indeed the dependency should be optional and default off.

pmarks:
@johanneskoester I made bio optional and fixed config.  Just need to set the version constraint for bio to match the new version.  What version is it going to be?

pmarks:
@johanneskoester - I've rebased this & made it optional against rust-bio >= 0.20. Should be ready now.

Unfortunately the From<Alignment> trait won't work because the user needs to be able to control whether the unaligned bases are hard-clipped or not, so we need an extra argument.

pmarks:
@johanneskoester - I've rebased this & made it optional against rust-bio >= 0.20. Should be ready now.

Unfortunately the From<Alignment> trait won't work because the user needs to be able to control whether the unaligned bases are hard-clipped or not, so we need an extra argument.

pmarks:
@johanneskoester  - how about this one?

johanneskoester:
Yes, for this, I have the following suggestion. Since some time, there is the [bio-types crate](https://github.com/rust-bio/rust-bio-types). The solution to solve the dependency issue is to move the `Alignment` type into bio-types, and then depend on bio-types from both rust-bio and rust-htslib.

pmarks:
@johanneskoester - great, I'm proceeding with that concept.  Can you give me write access to rust-bio-types so I can push a branch?  We'll then need to release that to crates.io and use the appropriate version

johanneskoester:
Sure, you have write access now. Thanks for working on this!

pmarks:
OK, once https://github.com/rust-bio/rust-bio-types/pull/2 is merged and published to crates.io, this can be merged.

pmarks:
@johanneskoester thanks for the bio-types release. This is ready to go now.

pmarks:
@johanneskoester  - added the docs.

PR #52, opened 2017-12-21
Add unset_* methods for bam record flags

pryvkin10x:
There's a case to be made for:
- unset_unmapped -> set_mapped
- unset_mate_unmapped -> set_mate_mapped

And maybe:
- unset_secondary -> set_primary
- unset_reverse -> set_forward

But I like the parallel language as is.


johanneskoester:
Thanks!

PR #58, opened 2018-02-11
serde support for Record objects

pmarks:
optional support for using Record objects with serde. Useful for integrating BAM data with Rust serde infrastructure.

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -10,6 +10,7 @@ use std::ops;
 use std::fmt;
 use std::error::Error;
 
+use serde::Serialize;
 use itertools::Itertools;
 
 use htslib;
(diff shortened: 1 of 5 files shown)


src/bam/record.rs
```diff
@@ -69,6 +69,56 @@ impl Clone for Record {
     }
 }
 
+impl PartialEq for Record {
+    fn eq(&self, other: &Record) -> bool {
```
johanneskoester: I think I would prefer a big lazy OR here: ` self.tid() != other.tid() || self.pos() != other.pos() || ...`

src/bam/record.rs
```diff
@@ -414,12 +481,8 @@ impl Record {
...
             }
         };
-
-        if ret < 0 {
-            Err(AuxWriteError::Some)
-        } else {
-            Ok(())
-        }
+        
+        Ok(())
```
johanneskoester: Why removing the error handling here?

src/bam/record.rs
pmarks: good call

src/bam/record.rs
pmarks: restored the error handling. In htslib 1.7, aux_append returns void, so we're going to need to drop this eventually.

PR #60, opened 2018-02-21
Extend derives for structs and enums

gavento:
* Add `Debug` for all structs and enums (even structures with only pointers -- it is still useful to auto-derive `Debug` for your struct containing those).
* Add `Clone` / `Copy` for structs & enums with simple data (no pointers, e.g. only `Vec`s, `Option`s, primitives, references).
* Add `PartialEq` / `Eq` and `Hash` to structs with clear semantics (not some seemingly binary-encoded data, there `PartialEq` might not match the data semantics)
* Add `Clone` for errors (not `Copy` in case you would like to extend some them with `String` messages or such)

Would you consider releasing a new version? (my code depends on copying `GenotypeAllele` and it would make my life and code distribution simpler) Thanks!


gavento:
Thanks!

johanneskoester:
New version is released.

PR #72, opened 2018-04-12
Extending support for setting thread count on I/O.

### LINKED ISSUE #69: Allow setting I/O threads everywhere htslib allows it.

holtgrewe:
fixes #69

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -124,6 +124,20 @@ impl Reader {
         let _self = unsafe { &*(data as *mut Self) };
         unsafe { htslib::bam_read1(_self.bgzf, record) }
     }
+
+    /// Activate multi-threaded BAM read support in htslib. This should permit faster
+    /// reading of large BAM files.
+    /// # Arguments
+    ///
+    /// * `n_threads` - number of extra background reader threads to use
+    pub fn set_threads(&mut self, n_threads: usize) -> Result<(), ThreadingError> {
+        let r = unsafe { htslib::bgzf_mt(self.bgzf, n_threads as i32, 256) };
+        if r != 0 {
+            Err(ThreadingError::Some)
+        } else {
+            Ok(())
+        }
+    }
 }
 
 
@@ -252,6 +266,20 @@ impl IndexedReader {
         match _self.itr {
             Some(itr) => itr_next(_self.bgzf, itr, record), // read fetched region
             None      => unsafe { htslib::bam_read1(_self.bgzf, record) } // ordinary reading
+        }
+    }
+
+    /// Activate multi-threaded BAM read support in htslib. This should permit faster
+    /// reading of large BAM files.
+    /// # Arguments
+    ///
+    /// * `n_threads` - number of extra background reader threads to use
+    pub fn set_threads(&mut self, n_threads: usize) -> Result<(), ThreadingError> {
+        let r = unsafe { htslib::bgzf_mt(self.bgzf, n_threads as i32, 256) };
+        if r != 0 {
+            Err(ThreadingError::Some)
+        } else {
+            Ok(())
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/bam/mod.rs
```diff
@@ -124,6 +124,29 @@ impl Reader {
...
+    /// reading of large BAM files.
+    ///
+    /// Setting `nthreads` to `0` does not change the current state.  Note that it is not
+    /// possible to set the number of background threads below `1` once it has been set.
+    ///
+    /// # Arguments
+    ///
+    /// * `n_threads` - number of extra background reader threads to use
+    pub fn set_threads(&mut self, n_threads: usize) -> Result<(), ThreadingError> {
+        if n_threads == 0 {
```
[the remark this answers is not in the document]
holtgrewe: Moving.

src/bam/mod.rs
```diff
@@ -388,10 +434,19 @@ impl Writer {
...
     /// # Arguments
     ///
-    /// * `n_threads` - number of background writer threads to use
+    /// * `n_threads` - number of extra background writer threads to use
     pub fn set_threads(&mut self, n_threads: usize) -> Result<(), ThreadingError> {
+        if n_threads == 0 {
+            // Do nothing if thread count is zero.  bgzf_mt() below would
+            // return -1 to indicate an error.
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
holtgrewe: OK

src/bcf/mod.rs
```diff
@@ -96,6 +96,29 @@ impl Reader {
...
+    /// Activate multi-threaded BCF read support in htslib. This should permit faster
+    /// reading of large BCF files.
+    ///
+    /// Setting `nthreads` to `0` does not change the current state.  Note that it is not
+    /// possible to set the number of background threads below `1` once it has been set.
+    ///
+    /// # Arguments
+    ///
+    /// * `n_threads` - number of extra background reader threads to use
+    pub fn set_threads(&mut self, n_threads: usize) -> Result<(), ThreadingError> {
```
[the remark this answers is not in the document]
holtgrewe: Ok

src/bcf/mod.rs
```diff
@@ -37,6 +37,25 @@ pub struct Reader {
 unsafe impl Send for Reader {}
 
 
+/// Implementation for `Reader::set_threads()` and `Writer::set_threads`.
+pub fn set_threads(hts_file: *mut htslib::htsFile, n_threads: usize)
+        -> Result<(), ThreadingError> {
+    assert!(n_threads > 0, "n_threads must be > 0");
+
+    if n_threads == 0 {
```
johanneskoester: Please remove the if since this is already handled by the assertion above.

src/tbx/mod.rs
```diff
@@ -204,6 +204,8 @@ impl Reader {
     ///
     /// * `n_threads` - number of extra background reader threads to use
     pub fn set_threads(&mut self, n_threads: usize) -> Result<(), ThreadingError> {
+        assert!(n_threads > 0, "n_threads must be > 0");
```
johanneskoester: Am I right that we can simply use  `bcf::set_threads` here? Maybe move it to utils.rs before.

src/bcf/mod.rs
holtgrewe: Of course.

src/tbx/mod.rs
holtgrewe: We would need a common `ThreadingError` type then...

holtgrewe: Can we maybe go forward with this patch anyway? I don't know so much about Rust error types and the best practice here (in python it's considered best practice to have an `exceptions` module for a library). Also, I'm planning to add an `IndexedReader` for VCF/BCF soon anyway and we can some more cleanup afterwards and consolidate the shared things? E.g., we should be able to centralize all rust code for parallel bgzf into one place including the thread pool stuff I am planning to add.

johanneskoester: Ok!

johanneskoester: The most standard solution for error handling is using the error-chain crate. I want to migrate error handling to that, but I did not find the time so far. See https://docs.rs/error-chain

PR #73, opened 2018-04-12
WIP Adding support for writing FILTER field.

### LINKED ISSUE #70: Add API for getting/setting FILTER column on BCF records

holtgrewe:
fixes #70 

- [x] writing `FILTER` field
- [ ] reading `FILTER` field
- [ ] reading `ID` field (whoops forgot in earlier PR)

### DIFF
--- a/src/bcf/record.rs
+++ b/src/bcf/record.rs
@@ -157,6 +157,92 @@ impl Record {
         }
     }
 
+    /// Set the given filters to the FILTER column.
+    ///
+    /// Note that the given strings must have been registered in the header and it will be
+    /// automatically converted to the corresponding integer value for being added.
+    ///
+    /// Setting an empty slice removes all filters.
+    ///
+    /// # Args
+    /// - `val` - The corresponding filter string value.
+    pub fn set_filters(&mut self, vals: &[Vec<u8>]) -> Result<(), FilterWriteError> {
+        let mut flt_ids: Vec<i32> = vals.iter().map(|val| {
+                let cval = ffi::CString::new(val.as_slice()).unwrap();
+                unsafe {
+                    htslib::bcf_hdr_id2int(
+                        self.header().inner,
+                        htslib::BCF_DT_ID as i32,
+                        cval.as_ptr() as *const i8,
+                    )
+                }
+            }
+        ).collect();
+        if unsafe { htslib::bcf_update_filter(
+            self.header().inner,
+            self.inner,
+            flt_ids.as_mut_ptr(),
+            flt_ids.len() as i32,
+        ) } == 0 {
+            Ok(())
+        } else {
+            Err(FilterWriteError::Some)
+        }
+    }
+
+    /// Add the given filter to the FILTER column.
+    ///
+    /// If `val` is `"PASS"` then all existing filters are removed first. If other than
+    /// `"PASS"`, then existing `"PASS"` is removed.
+    ///
+    /// Note that the given string must have been registered in the header and it will be
+    /// automatically converted to the corresponding integer value for being added.
+    ///
+    /// # Args
+    /// - `val` - The corresponding filter string value.
+    pub fn push_filter(&mut self, val: &[u8]) -> Result<(), FilterWriteError> {
+        let flt_id = unsafe { htslib::bcf_hdr_id2int(
+            self.header().inner,
+            htslib::BCF_DT_ID as i32,
+            ffi::CString::new(val).unwrap().as_ptr() as *const i8,
+        ) };
+        if unsafe { htslib::bcf_add_filter(
+            self.header().inner,
+            self.inner,
+            flt_id
+        ) } == 0 {
+            Ok(())
+        } else {
+            Err(FilterWriteError::Some)
+        }
+    }
+
+    /// Remove the given filter from the FILTER column.
+    ///
+    /// Note that the given string must have been registered in the header and it will be
+    /// automatically converted to the corresponding integer value for being added.
+    ///
+    /// # Args
+    /// - `val` - The corresponding filter string value.
+    /// - `pass_on_empty` - Set to "PASS" when removing the last value.
+    pub fn remove_filter(&mut self, val: &[u8], pass_on_empty: bool) -> Result<(), FilterWriteError> {
+        let flt_id = unsafe { htslib::bcf_hdr_id2int(
+            self.header().inner,
+            htslib::BCF_DT_ID as i32,
+            ffi::CString::new(val).unwrap().as_ptr() as *const i8,
+        ) };
+        if unsafe { htslib::bcf_remove_filter(
+            self.header().inner,
+            self.inner,
+            flt_id,
+            pass_on_empty as i32,
+        ) } == 0 {
+            Ok(())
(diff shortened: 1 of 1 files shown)


holtgrewe:
OK, this is read for review now.

src/bcf/record.rs
```diff
@@ -157,6 +157,96 @@ impl Record {
...
+    /// Set the given filters to the FILTER column.
+    ///
+    /// Note that the given strings must have been registered in the header and it will be
+    /// automatically converted to the corresponding integer value for being added.
+    ///
+    /// Setting an empty slice removes all filters.
+    ///
+    /// # Args
+    /// - `val` - The corresponding filter string value.
+    pub fn set_filters(&mut self, vals: &[Vec<u8>]) -> Result<(), FilterWriteError> {
```
johanneskoester: I wonder if we should allow to set filters by id, and get the ID first. I.e., expose a method `bcf::HeaderView::id(&[u8]) -> bcf::header::ID`. And then have here `set_filters(&mut self, vals: &[bcf::header::ID])`. What do you think? This also guarantees that only available IDs are provided. In a way, this supports the Rust paradigm of trying to fail at compile time already.

holtgrewe: Hm, think we should offer both an integer-based and a convenience interface. However, I don't see how this connects to failing at compile time. The integer-based API will allow for handling individual string-to-id resolution errors (which is useful) but nothing more.

holtgrewe: After thinking about this some more, I agree that we should rely on bcftools integers here. Adjusting accordingly.

johanneskoester: I meant that since the method `HeaderView::id` is the only one to get `bcf::header::ID`, the API kind of enforces that user cannot easily put an invalid ID into set_filters.

holtgrewe:
What's missing for this to move forward?

src/bcf/header.rs
```diff
@@ -180,6 +180,49 @@ impl HeaderView {
 
         Ok((_type, length))
     }
+
+    /// Convert string ID (e.g., for a `FILTER` value) to its numeric identifier.
+    pub fn id2int(&self, id: &[u8]) -> Result<u32, IdError> {
```
johanneskoester: Please return a newtype `bcf::header::ID` here, see: https://doc.rust-lang.org/stable/rust-by-example/generics/new_types.html. For implementing the newtype, use the crate https://docs.rs/newtype_derive. This allows you to inherit methods from u32.

src/bcf/header.rs
```diff
@@ -180,6 +180,49 @@ impl HeaderView {
...
+            ) {
+                -1 => Err(IdError::UnknownID(str::from_utf8(id).unwrap().to_owned())),
+                i => Ok(i as u32),
+            }
+        }
+    }
+
+    /// Convert integer representing an identifier (e.g., a `FILTER` value) to its string
+    /// name.bam
+    pub fn int2id(&self, id: i32) -> Vec<u8> {
```
johanneskoester: Take above newtype here, call this function `id_to_name`.

src/bcf/header.rs
johanneskoester: Call this function `name_to_id`.

src/bcf/header.rs
```diff
@@ -180,6 +180,49 @@ impl HeaderView {
...
+        let key = unsafe { ffi::CStr::from_ptr(
+            (*(*self.inner).id[htslib::BCF_DT_ID as usize].offset(id as isize)).key) };
+        key.to_bytes().to_vec()
+    }
+
+    /// Convert string sample name to its numeric identifier.
+    pub fn sample2int(&self, id: &[u8]) -> Result<u32, SampleError> {
```
johanneskoester: Call this function `sample_to_id`.

src/bcf/header.rs
```diff
@@ -180,6 +180,49 @@ impl HeaderView {
...
+                ffi::CString::new(id).unwrap().as_ptr() as *const i8
+            ) {
+                -1 => Err(SampleError::UnknownSample(str::from_utf8(id).unwrap().to_owned())),
+                i => Ok(i as u32),
+            }
+        }
+    }
+
+    /// Convert integer representing an contig to its name.
+    pub fn int2sample(&self, id: i32) -> Vec<u8> {
```
johanneskoester: `id_to_sample`

src/bcf/record.rs
```diff
@@ -157,6 +157,52 @@ impl Record {
         }
     }
 
+    /// Set the given filters IDs to the FILTER column.
+    ///
+    /// Setting an empty slice removes all filters.
+    ///
+    /// # Args
+    /// - `val` - The corresponding filter string value.
+    pub fn set_filters(&mut self, flt_ids: &[i32]) {
```
johanneskoester: Signature should be `flt_ids: &[bcf::header::ID]`.

src/bcf/record.rs
```diff
@@ -157,6 +157,52 @@ impl Record {
...
+    }
+
+    /// Add the given filter to the FILTER column.
+    ///
+    /// If `val` corresponds to `"PASS"` then all existing filters are removed first. If other than
+    /// `"PASS"`, then existing `"PASS"` is removed.
+    ///
+    /// # Args
+    /// - `val` - The corresponding filter ID value.
+    pub fn push_filter(&mut self, flt_id: u32) {
```
johanneskoester: Signature should be `flt_id: bcf::header::ID`.

src/bcf/record.rs
```diff
@@ -157,6 +157,52 @@ impl Record {
...
+            htslib::bcf_add_filter(self.header().inner, self.inner, flt_id as i32);
+        }
+    }
+
+    /// Remove the given filter from the FILTER column.
+    ///
+    /// # Args
+    /// - `val` - The corresponding filter ID.
+    /// - `pass_on_empty` - Set to "PASS" when removing the last value.
+    pub fn remove_filter(&mut self, flt_id: u32, pass_on_empty: bool) {
```
johanneskoester: Signature should be `flt_id: bcf::header::ID, pass_on_empty: bool`.

src/bcf/header.rs
holtgrewe: OK

src/bcf/header.rs
holtgrewe: OK

src/bcf/header.rs
holtgrewe: OK

holtgrewe:
OK, introduced `Id` type.

src/bcf/header.rs
```diff
@@ -14,6 +14,26 @@ use htslib;
...
+        NewtypeDiv(*),
+        PartialEq,
+        PartialOrd,
+        Copy,
+        Clone,
+        Debug,
+        Default
+    )]
+    pub struct Id(pub u32);
+}
```
johanneskoester: I think I would probably not derive Add, Sub, Mul, Div and Default, as there is no need for these, and they could only be used to generate invalid ids. However, I can also change this after merging

Issue #68, opened 2018-04-12
Add wrapper for bcf_srs_t, bcf_sr_seek, etc.


holtgrewe:
This part of htslib allows indexed reading of BCF/VCF (transparently) and also iteration in multiple files at the same time.

johanneskoester:
Done as well.

PR #74, opened 2018-04-16
Starting bcf::IndexedReader

holtgrewe:
Partially adresses #68. However, only streaming over one file is currently implemented to give a `bcf::IndexedReader` interface.

Note that I somewhere made a mistake so the tests crash with an invalid memory access error.

I have to admit that I don't have an idea how to approach this with Rust tooling and would appreciate any help!


src/bcf/mod.rs
```diff
@@ -92,10 +99,17 @@ impl Reader {
...
-    pub fn records(&mut self) -> Records {
+    fn records(&mut self) -> Records<Self> {
         Records { reader: self }
     }
+
+    fn header(&self) -> &HeaderView {
+        &self.header
+    }
+
+    fn empty_record(&self) -> Record {
```
johanneskoester: You should be able to move the implementation of empty_record into the Trait, by using the Read::header() method, right?

src/bcf/mod.rs
```diff
@@ -108,6 +122,125 @@ impl Drop for Reader {
...
+    }
+}
+
+
+impl Read for IndexedReader {
+    fn read(&mut self, record: &mut record::Record) -> Result<(), ReadError> {
+        match unsafe { htslib::bcf_sr_next_line(self.inner) } {
+            0 => Err(ReadError::NoMoreRecord),
+            i => {
+                assert!(i > 0, "Must not be negative");
```
johanneskoester: Can't this be negative in case of an error when reading the file? I guess we would like to return a `ReadError` variant in such a case, rather than having a failing assertion.

src/bcf/mod.rs
holtgrewe: OK

src/bcf/mod.rs
holtgrewe: Actually, the code counts the number of readers for which a record is found... The API is a bit weird in this area.

holtgrewe:
OK, pushed some fixes. I would propose to change the `bcf::Record` object as documented in the code at some later point when we have a more complete implementation of the synced reader and have a better feeling for that part of the API.

src/bcf/mod.rs
holtgrewe: Arg, then I need to expose the `Rc<HeaderView>` member. Is this OK?

src/bcf/mod.rs
johanneskoester: And how is an error indicated?

holtgrewe: I don't know. Here is the usage example from the docs:

```
    Example of usage:

        bcf_srs_t *sr = bcf_sr_init();
        bcf_sr_set_opt(sr, BCF_SR_PAIR_LOGIC, BCF_SR_PAIR_BOTH_REF);
        bcf_sr_set_opt(sr, BCF_SR_REQUIRE_IDX);
        for (i=0; i<nfiles; i++)
            bcf_sr_add_reader(sr,files[i]);
        while ( bcf_sr_next_line(sr) )
        {
            for (i=0; i<nfiles; i++)
            {
                bcf1_t *line = bcf_sr_get_line(sr,i);
                ...
            }
        }
        if ( sr->errnum ) error("Error: %s\n", bcf_sr_strerror(sr->errnum));
        bcf_sr_destroy(sr);
```

holtgrewe: gah, it's right there...

johanneskoester: Ah, looks like an error leads to termination of the loop (no more record) and then you can find an error code in `sr->errnum`? So, this means here that in the `0` case you would have to check `sr->errnunm` as well.

src/bcf/mod.rs
```diff
@@ -22,10 +22,41 @@ pub use bcf::buffer::RecordBuffer;
...
+    ///
+    /// # Arguments
+    /// * record - an empty record, that can be created with `bcf::Reader::empty_record`.
+    fn read(&mut self, record: &mut record::Record) -> Result<(), ReadError>;
+
+    /// Return an iterator over all records of the VCF/BCF file.
+    fn records(&mut self) -> Records<Self>;
+
+    /// Return the header.
+    fn header(&self) -> Rc<HeaderView>;
```
johanneskoester: Why not returning `&HeaderView`?

johanneskoester: Ah, I think now I understand the problem with empty_record. In that case, I think it is better to keep the API clean, return &HeaderView, and leave the implementation of empty_record to the implementing structs.

holtgrewe: Because `bcf::record::Record::new` wants a `Rc<HeaderView>`...

src/bcf/mod.rs
```diff
@@ -132,6 +151,149 @@ impl Drop for Reader {
...
+                    htslib::bcf_copy(
+                        record.inner,
+                        *(*(*self.inner).readers.offset(0)).buffer.offset(0));
+                }
+
+                record.set_header(self.header.clone());
+
+                match self.current_region {
+                    Some((rid, _start, end)) => {
+                        if record.rid().is_some() && rid == record.rid().unwrap() &&
```
johanneskoester: One last thing. Why is this necessary? I thought the fetch already fetches the correct region?

holtgrewe: It only jumps to the start of the region...

johanneskoester: ok!

PR #76, opened 2018-04-17
BCF header records

holtgrewe:
This works but I'm unsure about the design being right.

@johanneskoester what do you think?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ lazy_static = "0.2"
 bitflags = "0.9"
 serde = { version = "^1", optional = true }
 regex = "0.2"
+linear-map = "1.2.0"
 
 [features]
 default = []
--- a/src/bcf/header.rs
+++ b/src/bcf/header.rs
@@ -192,6 +192,7 @@ impl Drop for Header {
 }
 
 /// A header record.
+#[derive(Debug)]
 pub enum HeaderRecord {
     /// A `FILTER` header record.
     Filter {
--- a/src/bcf/header.rs
+++ b/src/bcf/header.rs
@@ -191,6 +191,22 @@ impl Drop for Header {
     }
 }
 
+/// A header record.
+pub enum HeaderRecord {
+    /// A `FILTER` header record.
+    Filter { key: String, key_value_pairs: Vec<(String, String)> },
+    /// An `INFO` header record.
+    Info { key: String, key_value_pairs: Vec<(String, String)> },
+    /// A `FORMAT` header record.
+    Format { key: String, key_value_pairs: Vec<(String, String)> },
+    /// A `contig` header record.
+    Contig { key: String, key_value_pairs: Vec<(String, String)> },
+    /// A structured header record.
+    Structured { key: String, key_value_pairs: Vec<(String, String)> },
+    /// A generic, unstructured header record.
+    Generic { key: String, value: String },
+}
+
 #[derive(Debug)]
 pub struct HeaderView {
     pub inner: *mut htslib::bcf_hdr_t,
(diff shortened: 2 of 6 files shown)


src/bcf/header.rs
```diff
@@ -240,6 +257,54 @@ impl HeaderView {
             (*(*self.inner).id[htslib::BCF_DT_SAMPLE as usize].offset(*id as isize)).key) };
         key.to_bytes().to_vec()
     }
+
+    /// Return structured `HeaderRecord`s.
+    pub fn header_records(&self) -> Vec<HeaderRecord> {
```
johanneskoester: I would return a HashMap here, which assigns keys to HeaderRecord.

src/bcf/header.rs
```diff
@@ -110,6 +110,23 @@ impl Drop for Header {
 }
 
 
+/// A header record.
+pub enum HeaderRecord {
+    /// A `FILTER` header record.
+    Filter { key: String, key_value_pairs: Vec<(String, String)> },
```
johanneskoester: I think these should be `<type> {key: String, values: HashMap<String, String>}`, except for Generic.

src/bcf/header.rs
johanneskoester: We can use LinearMap instead, there are only a few keys involved here. See https://docs.rs/linear-map

src/bcf/header.rs
johanneskoester: Wait, this should be LinearMap.

src/bcf/header.rs
johanneskoester: Right. And moreover there might be duplicate keys I think. Returning a Vector is fine.

holtgrewe: Hm, so should I just rebase or perform any changes?

holtgrewe: OK, resolving the merge conflicts now FWIW, would have to be done anyway.

holtgrewe:
🎉 @johanneskoester The test runs through, the code compiles and is properly formatted. Please advise on any further need for change.

holtgrewe:
@johanneskoester ping

src/bcf/header.rs
```diff
@@ -191,6 +191,38 @@ impl Drop for Header {
     }
 }
 
+/// A header record.
+#[derive(Debug)]
+pub enum HeaderRecord {
+    /// A `FILTER` header record.
+    Filter {
+        key: String,
+        key_value_pairs: Vec<(String, String)>,
```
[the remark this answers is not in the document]
holtgrewe: OK, did so, let's see what the CI build says.

holtgrewe:
@johanneskoester Well, this runs through in principle but doctests now also fail on stable... Otherwise, I incorporated your changes. What now?

Cargo.toml
```diff
@@ -35,4 +36,4 @@ pretty_assertions = "0.5.1"
 
 [build-dependencies]
 fs-utils = "1.0"
-bindgen = "0.35.0"
+bindgen = "0.36.0"
```
johanneskoester: Without the trailing zero, just `0.36`? Then we automatically get patch releases.

holtgrewe:
@johanneskoester No it's not. And now?

johanneskoester:
This is either a bug in rust or a behavior change in rust and a bug in rust-bindgen. I have created an issue: https://github.com/rust-lang-nursery/rust-bindgen/issues/1313

johanneskoester:
Can you update to the latest master of rust-htslib? I have temporarily pinned rust to 1.25

johanneskoester:
Yay! Thanks a lot!

PR #77, opened 2018-04-18
Adding bcf::Record::has_filter().

holtgrewe:


src/bcf/record.rs
```diff
@@ -131,6 +131,17 @@ impl Record {
         self.inner_mut().pos = pos;
     }
 
+    /// Return whether the filter with the given ID is set.
+    pub fn has_filter(&self, id: &[u8]) -> bool {
```
johanneskoester: Shouldn't this be `&[bcf::header::ID]`?

holtgrewe: Actually, the C API exposes `char const *` here. What do we want?

holtgrewe:
Should be resolved by #81.

PR #85, opened 2018-05-06
Adding removal of records from BCF header.

holtgrewe:


src/bcf/header.rs
```diff
@@ -35,21 +35,35 @@ pub struct Header {
...
     }
 
+    /// Create a new `Header` using the given `HeaderView as the template.
+    ///
+    /// After construction, you can modify the header independently from the template `header`.
+    ///
+    /// # Arguments
+    ///
+    /// - `header` - The `HeaderView` to use as the template.
     pub fn with_template(header: &HeaderView) -> Self {
```
johanneskoester: THis is a good opportunity to rename the function to `from_template`, following the Rust API guidelines for constructors.

src/bcf/header.rs
```diff
@@ -35,21 +35,35 @@ pub struct Header {
...
     }
 
+    /// Create a new `Header` using the given `HeaderView` as as template, but subsetting to the
+    /// given `samples`.
+    ///
+    /// # Arguments
+    ///
+    /// - `header` - The `HeaderView` to use for the template.
+    /// - `samples` - A slice of byte-encoded (`[u8]`) sample names.
     pub fn subset_template(header: &HeaderView, samples: &[&[u8]]) -> Result<Self, SubsetError> {
```
johanneskoester: This should be `from_template_subset`

holtgrewe: Changed  and rebased.

src/bcf/header.rs
holtgrewe: done

PR #87, opened 2018-05-17
add a 'chunked' iterator, Header --> HashMap, make cigar_len() public

pmarks:
Various API additions that we've found useful.


pmarks:
@johanneskoester - requested changes made. I tested with regex 0.2 - 1.0. since that's a common crate, does it make sense to try a support a wider version range?

johanneskoester:
Don't know. Do you know an example where this has been done?

pmarks:
I don't see any mention of it here: https://rust-lang-nursery.github.io/api-guidelines/checklist.html

but given that it compiles w/ versions up to 1.0, we might as well have it be >=0.2

johanneskoester:
Why not just switching to 1.0. that should be fine.

johanneskoester:
That should be not a problem, as rust compiles stuff statically. It might just increase the binary size, because the regex crate is compiled in twice.

johanneskoester:
I think your commit is not yet pushed?

pmarks:
@johanneskoester - sorry, this dropped off my radar.  I think this is ready now.

PR #88, opened 2018-06-07
Fixing bug with reading BCF filter fields

holtgrewe:


johanneskoester:
Good catch! Thank you!

PR #89, opened 2018-06-08
Fixed bug in missing unpack when retrieving BCF record id.

holtgrewe:


src/bcf/record.rs
```diff
@@ -157,6 +157,12 @@ impl Record {
     ///
     /// When empty, returns `b".".to_vec()`.
     pub fn id(&self) -> Vec<u8> {
+        // TODO: do we have to use `&mut self` here because of the unpack?!
```
johanneskoester: Yes, I think so. Otherwise, we could get a race condition. Would it make sense to put the unpack into the the constructor? Or does it cause a big overhead?

holtgrewe:
I'm uncertain here.

Maybe merge now and add issue to clean this up?

--
Dr. Manuel Holtgrewe, Dipl.-Inform.
Bioinformatician
Core Unit Bioinformatics – CUBI
Berlin Institute of Health / Max Delbrück Center for Molecular Medicine in the Helmholtz Association / Charité – Universitätsmedizin Berlin

Visiting Address: Invalidenstr. 80, 3rd Floor, Room 03 028, 10117 Berlin
Postal Address: Chariteplatz 1, 10117 Berlin

E-Mail: manuel.holtgrewe@bihealth.de
Phone: +49 30 450 543 607
Fax: +49 30 450 7 543 901
Web: cubi.bihealth.org  www.bihealth.org  www.mdc-berlin.de  www.charite.de

Am 10.06.2018 2:29 nachm. schrieb Johannes Köster <notifications@github.com>:

@johanneskoester commented on this pull request.

________________________________

In src/bcf/record.rs<https://github.com/rust-bio/rust-htslib/pull/89#discussion_r194255730>:

> @@ -157,6 +157,12 @@ impl Record {
     ///
     /// When empty, returns `b".".to_vec()`.
     pub fn id(&self) -> Vec<u8> {
+        // TODO: do we have to use `&mut self` here because of the unpack?!


Yes, I think so. Otherwise, we could get a race condition. Would it make sense to put the unpack into the the constructor? Or does it cause a big overhead?

johanneskoester:
We can't merge if we know of the race condition. I think it should be fine to add the unpack to the constructor.

holtgrewe:
Well, we merged with this TODO previously ;) I can look into this tomorrow, however.

holtgrewe:
OK, I believe this should do it.

holtgrewe:
@johanneskoester OK, unpacking is done on reading and construction now.

johanneskoester:
Awesome, thank you!

PR #92, opened 2018-06-18
Initial implementation of bcf::SyncedReader

holtgrewe:
Caution: work in progress, do not merge.


holtgrewe:
Hi @johanneskoester, this is ready for review now.

holtgrewe:
The failures are outside my patch (but also, I cannot change formatting with `cargo fmt` for some reason). They were previously introduced by you @johanneskoester, I believe ;)

johanneskoester:
This is now fixed in master.

holtgrewe:
OK, le'ts see whether this does the trick.

holtgrewe:
@johanneskoester OK, this is ready for review now.

src/bcf/mod.rs
```diff
@@ -303,6 +303,158 @@ impl Drop for IndexedReader {
...
+
+    /// Enable or disable requiring of index
+    pub fn set_require_index(&mut self, do_require: bool) {
+        unsafe {
+            (*self.inner).require_index = if do_require { 1 } else { 0 };
+        }
+    }
+
+    /// Set the given bitmask of values from `sr_pairing` module.
+    pub fn set_pairing(&mut self, bitmask: u32) {
```
johanneskoester: I think it would be nicer to have an enum for the different pairings and hide this bitmask details inside `set_pairing`.

src/bcf/mod.rs
```diff
@@ -303,6 +303,158 @@ impl Drop for IndexedReader {
...
+                );
+            }
+            Some(record)
+        } else {
+            None
+        }
+    }
+
+    /// Return header from the given reader.
+    pub fn get_header(&self, idx: u32) -> &HeaderView {
```
johanneskoester: Rust style is without the `get_`.

src/bcf/mod.rs
```diff
@@ -303,6 +303,158 @@ impl Drop for IndexedReader {
...
+    pub fn has_line(&self, idx: u32) -> bool {
+        if idx >= self.reader_count() {
+            panic!("Invalid reader!");
+        } else {
+            unsafe { (*(*self.inner).has_line.offset(idx as isize)) != 0 }
+        }
+    }
+
+    /// Return record from the given reader, if any.
+    pub fn get_record(&self, idx: u32) -> Option<Record> {
```
johanneskoester: Rust style is without the `get_`.

src/bcf/mod.rs
holtgrewe: I think about any combination of bitmasks is possible, so I would argue against this. When the dust settles a bit on Rust creates, we might rely on an external crate here, but I would refrain from that until there crystalizes a best practice for bit masks.

src/bcf/mod.rs
holtgrewe: Can do.

src/bcf/mod.rs
holtgrewe: Can do.

src/bcf/mod.rs
johanneskoester: Ah, sorry, now I get it. Indeed, I agree.

src/bcf/mod.rs
```diff
@@ -303,6 +303,158 @@ impl Drop for IndexedReader {
     }
 }
 
+/// This module contains bitmask constants for `SyncedReader`.
```
johanneskoester: Maybe put the entire symmetric reader code in a submodule of bcf and call `sr_paring` `pairing` instead.

holtgrewe: Hm, although this is not Python, I like the "flat is better than nested" theme ;) If you think it's better style, I can do this.

johanneskoester: Yes. The indicator here is that you have to use sr_pairings for the module with the constants. Further, it is quite some complexity added to bcf, that is not necessarily part of it. Hence I think it makes sense to add it as a submodule.

src/bcf/mod.rs
holtgrewe: 👍

src/bcf/mod.rs
holtgrewe: 👍

src/bcf/mod.rs
holtgrewe: 👍

src/bcf/mod.rs
holtgrewe: 👍

holtgrewe:
OK, adjusted code according to your remarks.

PR #93, opened 2018-06-19
make bzip2-/lzma-support package features

nlhepler:
This propagates the support of bzip2 and lzma up to cargo, in case the user/deployment environment does not support one or both of these libraries.


johanneskoester:
Interesting! Would this also work for zlib?

nlhepler:
@johanneskoester --I haven't tried, but I don't think so--zlib dependence is baked deep into htslib, whereas bzip2 and lzma have explicit options inside the configure script disabling their support. You can see evidence for this here: https://github.com/samtools/htslib/blob/develop/configure.ac#L78, and in bgzf.h which does _not_ wrap `#include <zlib.h>` with a compile-time definition.

johanneskoester:
May I ask you to merge in latest master?

PR #104, opened 2018-07-31
Update htslib to v1.9

juliangehring:
Updates the htslib submodule from v1.8 to v1.9 (see https://github.com/samtools/htslib/releases for a list of the relevant changes). All tests complete successfully with rust v1.27.2 and v1.29.0-nightly.


johanneskoester:
Thanks!

Issue #80, opened 2018-04-29
Static compilation possible?


wwood:
Hi,

I'm trying to create a static binary for a project of mine https://github.com/wwood/CoverM but i'm running into an issue: 
```
$ rustup target add x86_64-unknown-linux-musl
...
$ cargo build --target x86_64-unknown-linux-musl --release
...
/usr/bin/ld: /home/ben/git/coverm/target/x86_64-unknown-linux-musl/release/deps/librust_htslib-f71d6303e993dcd2.rlib(sam_header.o): undefined reference to symbol '__printf_chk@@GLIBC_2.3.4'
```
Is it possible to get a static compile of a project using rust-htslib? Am I doing something incorrectly?
ta, ben

johanneskoester:
Static building is fine. I do that all the time. However, HTSlib needs glibc >= 2.3.4. You are trying to build on musl instead. I don't know if that is supported. I suggest asking the htslib guys.

johanneskoester:
Closing for now, feel free to reopen if you disagree or new issues arise.

wwood:
Thanks, I would be happy to use glibc. However, after a frustrating amount of internet searching I'm finding it difficult to work out how, exactly. Would you mind sharing your secret? I'm looking to build an executable.

Apologies if this is a naive question, I just figure you might have a solution easily at hand.
Thanks, ben

johanneskoester:
glibc should be the default on most linux distributions. So, if you e.g. run on Ubuntu or Fedora, it will use a new enough glibc automatically. Unfortunately, if your system uses musl, you might need to do the build inside a container or a virtual machine, or try to figure out how to switch to glibc on your system.

wwood:
Right, glibc is the default on my (Ubuntu/GNU Guix) system. I installed the musl toolchain with rustup specifically for the purpose of static compilation, following this advice:

> Static linking is supported via an alternative libc, musl.

https://doc.rust-lang.org/1.12.1/book/advanced-linking.html#static-linking

Compiling with the musl toolchain gives a static binary by default, where the glibc toolchain does not (I tried using `ldd` and it failed to run on another system). Do you have any suggestions on how to ask for a static compilation with the glibc toolchain, or some other solution?

Thanks, ben

johanneskoester:
[...]

PR #111, opened 2018-09-04
feat: add musl target support

zitsen:
Changes:
- explicitly declare deps for `bzip2` and `lzma` feature.
- add musl target support using `cc` crate.

Though I choose to link native deps via the `bzip2-sys` and `lzma-sys` crates,
note that the deps could be done with cargo's `target` selector.
Let me know if you do not want to link it by default.

/cc #80

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,9 @@ documentation = "https://docs.rs/rust-htslib"
 
 [dependencies]
 libc = "0.2"
+libz-sys ="1.0"
+bzip2-sys = { version = "0.1.6", git = "https://github.com/alexcrichton/bzip2-rs.git", optional = true }
+lzma-sys = { version = "0.1.10", optional = true }
 itertools = "0.6"
 quick-error = "1.2"
 newtype_derive = "0.1"
@@ -28,8 +31,8 @@ bio-types = ">=0.1.1"
 
 [features]
 default = ["bzip2", "lzma"]
-bzip2 = []
-lzma = []
+bzip2 = ["bzip2-sys"]
+lzma = ["lzma-sys"]
 
 [dev-dependencies]
 tempdir = "0.3"
@@ -40,3 +43,4 @@ pretty_assertions = "0.5.1"
 [build-dependencies]
 fs-utils = "1.0"
 bindgen = "0.36"
+cc = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ documentation = "https://docs.rs/rust-htslib"
 
 [dependencies]
 libc = "0.2"
-libz-sys ="1.0"
+libz-sys = "1.0"
 bzip2-sys = { version = "0.1.6", git = "https://github.com/alexcrichton/bzip2-rs.git", optional = true }
 lzma-sys = { version = "0.1.10", optional = true }
 itertools = "0.6"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -13,6 +13,9 @@ documentation = "https://docs.rs/rust-htslib"
 
 [dependencies]
 libc = "0.2"
+libz-sys = "1.0"
+bzip2-sys = { version = "0.1.6", git = "https://github.com/alexcrichton/bzip2-rs.git", optional = true }
```
johanneskoester: Not sure I understand this. I thought these sys crates are just the bindings, but we don't need bindings here do we?

build.rs
```diff
@@ -39,17 +48,28 @@ fn main() {
     if !use_bzip2 {
         let bzip2_patterns = vec!["s/ -lbz2//", "/#define HAVE_LIBBZ2/d"];
         sed_htslib_makefile(&out, &bzip2_patterns, "bzip2");
+    } else if let Ok(inc) = env::var("DEP_BZIP2_ROOT")
```
johanneskoester: Why are these additional env vars needed?

Cargo.toml
johanneskoester: Do they also ship the libs?

johanneskoester:
Looks really nice! just a few questions of somebody not so experienced with C compilation.

build.rs
johanneskoester: Do they come from the bzip2-sys crate?

zitsen: that's so sad I mistakenly deleted the commonts :cry: , maybe you could see them in email?

zitsen:
There's something to do following `libz-sys` style:

1. Windows support should be dealt with, test should be passed (via appveyor - it's upon to you).
2. More targets should be tested(this is optional)
3. [Trust](https://github.com/japaric/trust) will help.
4. filling build.rs

I'm glad to help after your done with appveyor(it's free).

johanneskoester:
Why would we need windows testing? As far as I know, htslib does not support windows, [see here](https://github.com/samtools/htslib/issues/86). Also, windows is not really relevant as a platform for the kind of bioinformatics analyses that can be done with htslib and rust-htslib.

zitsen:
ignore it and make your decision then.

johanneskoester:
Yes, I think libz-sys support would be nice. Do you want to add it in this PR?

Cargo.toml
zitsen: libz-sys is here

zitsen:
did you mean add libz-sys dep? it has already been included.

or libz-sys like build scripts for htslib???

maybe my english is urgly....

johanneskoester:
I just meant adding at a dependency and using it. I was confused by the fact that there is no feature for it, and apparently it is not used in the build.rs. Is that intentional?

zitsen:
see [`libz-sys` added as a dependent here](https://github.com/rust-bio/rust-htslib/pull/111/files#diff-80398c5faae3c069e4e6aa2ed11b28c0R16). in common, there's no need to declare it in `build.rs`, but do [declare it in `lib.rs`](https://github.com/rust-bio/rust-htslib/pull/111/files#diff-b4aea3e418ccdb71239b96952d9cddb6R102), cargo link it automatically. bzip2 and lzma is declared as optional because of feature deps, but libz is always required.

any else?

johanneskoester:
Thank you!

PR #115, opened 2018-11-16
increase version range of bindgen

pmarks:
Allow a wider range of bindgen versions. Bindgen depends on clang-sys which has a native dependency, therefore only one version can be present in a build.  Pinning to a single bindgen version makes it difficult to combine rust-htslib with other crates that use bindgen. 

I've tested that bindgen 0.37 - 0.43 pass the rust-htslib tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
       - libcurl4-openssl-dev
       - libelf-dev
       - libdw-dev
+      - libbz2-dev
       - binutils-dev
       - libiberty-dev
       - cmake
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,5 +42,5 @@ pretty_assertions = "0.5.1"
 
 [build-dependencies]
 fs-utils = "1.0"
-bindgen = "0.36"
+bindgen = ">=0.37"
 cc = "1.0"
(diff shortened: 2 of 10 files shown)


pmarks:
@johanneskoester I fixed the rustfmt problem.  

There's a problem finding bzlib.h in the htslib build which seems to pre-date this PR:
https://travis-ci.org/rust-bio/rust-htslib/builds

libbz2-dev is installed, so I'm not sure how to coax it to find that header...

johanneskoester:
Thanks for letting me know. I will have a look tomorrow.

johanneskoester:
@zitsen musl builds are suddenly not working anymore. Seems like the bzlib header cannot be found, although I thought that the bzip2-sys crate is already providing it as a fall back as you said, right? Any idea what's wrong?

johanneskoester:
@pmarks in order to proceed here, feel free to disable musl builds for now.

zitsen:
@johanneskoester @pmarks that is an bzip2-sys crate version error, i've filed an issue in bzip2-sys repo alexcrichton/bzip2-rs#36 .

at current time, please fix the `bzip2-sys` to git version like my [original pr does](https://github.com/rust-bio/rust-htslib/pull/111/files#diff-80398c5faae3c069e4e6aa2ed11b28c0R17)

johanneskoester:
We can't do a release of rust bio as long as there is a git version in the dependencies. Maybe an older bzip-sys version would work?

PR #117, opened 2018-12-02
bam: fix l_extranul processing when deserializing

nlhepler:
Turns out if you don't do this, you may see between 1-3 extra null bytes at the end of your record qname because of alignment padding.

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -240,14 +240,18 @@ impl Record {
         self.inner_mut().core.isize = insert_size;
     }
 
-    fn qname_len(&self) -> usize {
+    fn qname_capacity(&self) -> usize {
         self.inner().core.l_qname as usize
+    }
+
+    fn qname_len(&self) -> usize {
+        // discount all trailing zeros (the default one and extra nulls)
+        self.qname_capacity() - 1 - self.inner().core.l_extranul as usize
     }
 
     /// Get qname (read name). Complexity: O(1).
     pub fn qname(&self) -> &[u8] {
-        // remove all trailing zeros (the default one and extra nulls)
-        &self.data()[..self.qname_len() - 1 - self.inner().core.l_extranul as usize]
+        &self.data()[..self.qname_len()]
     }
 
     /// Set the variable length data buffer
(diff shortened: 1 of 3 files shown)


src/bam/record_serde.rs
```diff
@@ -6,6 +6,12 @@ use serde::{Serialize, Serializer};
 
 use bam::record::Record;
 
+fn fix_l_extranul(mut rec: Record) -> Record {
```
johanneskoester: Maybe work on a mutable reference of record, and don't return it?

nlhepler:
@johanneskoester, I also changed PartialEq to take this into account, added it to the serde tests, and altered the travis.yml to test all features in addition to no-default-features.

src/bam/record.rs
```diff
@@ -81,6 +81,7 @@ impl PartialEq for Record {
             && self.mpos() == other.mpos()
             && self.insert_size() == other.insert_size()
             && self.data() == other.data()
+            && self.inner().core.l_extranul == other.inner().core.l_extranul
```
johanneskoester: Extranul does not really change anything in terms of semantics, right? I wonder if we rather should consider two records the same regardless of the amount of padding.

nlhepler: If I were to recommend any changes to the API, qname_len() should probably be modified to take l_extranul into account (it would make more sense--someone is probably more interested in the (&str-like) bytes of the header instead of the &CStr + [0-3] bytes of (extra) null-padding.

johanneskoester: Yes, I agree.

johanneskoester: If you like, please do that in this PR.

johanneskoester: Excellent. Now I think it would make sense to also remove the l_extranul check above, wouldn't it?

nlhepler: I don't think so--the l_extranul comparison is still required to guarantee that qname() yields the same output between objects otherwise identical. All other non-variable data are either implicitly captured by the data() comparison or are explicitly represented--all except l_extranul.

johanneskoester:
Ok, thanks a lot!

src/bam/record.rs
nlhepler: I can remove it (I don't think it looks very pleasant either), but then e.g. test_bincode() and test_serde_json() may pass, even if l_extranul handling is broken.

Alternatively, I could implement Eq for bam1_core_t instead?

Something else I now notice, looking more closely--I think we should take care to align the beginning of the cigar as intended (the whole purpose behind l_extranul). If you agre, feel free to make an issue and assign it to me.

johanneskoester: Yes, implementing Eq for bam1_core_t makes more sense indeed.

Issue #116, opened 2018-11-19
CRAM support


juliangehring:
Out of curiosity, are there plans to add a high-level interface for reading/writing CRAM files?

johanneskoester:
High level in what sense? The current bam module of rust-htslib already supports reading and writing CRAM I think. Or am I missing something?

juliangehring:
How would I provide the [reference genome file](http://www.htslib.org/doc/samtools.html#REFERENCE_SEQUENCES) with the current interface? I don't see where this would need to go with the current `bam::Reader::from_path(&"test/test.cram").unwrap();` - but maybe I'm missing something here.

juliangehring:
Let's use a [small, public data set](https://github.com/Illumina/strelka/tree/v2.9.x/src/demo/data) to make it more concrete:
The BAM files are already there, we can build the matching CRAM files with e.g.
```sh
samtools view -C -T demo20.fa -o NA12891_demo20.cram NA12891_demo20.bam
```
With
```rust
use rust_htslib::bam;   // rust_htslib v0.22.0
let mut bam = bam::Reader::from_path(&path).unwrap();
```
reading the BAM file `NA12891_demo20.bam` works fine, but trying to read the CRAM file fails with:

> [W::bam_hdr_read] EOF marker is absent. The input is probably truncated
[E::bam_hdr_read] Invalid BAM binary header
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Invalid', libcore/result.rs:1009:5

johanneskoester:
Thanks for letting me know. You are right then! I always assumed the bam reader is format-agnostic. So, yes, in that case, adding CRAM support is definitely an intended feature now. However, I must admit that my schedule is so tight that I would need help in order to get this done quickly. Hence I cannot promise a timeline. However, I am happy to accept any PRs.

juliangehring:
Thanks, it would be neat to have cram support in the library as well. I'm happy to help where I can, but my knowledge of the htslib internals is very limited - not sure if I can be of big help here.

juliangehring:
#123 brought in all the basic functionality for reading and writing CRAM files.
Thanks @jfarek!

PR #123, opened 2019-03-20
Implement CRAM read and write in bam module (#116)

jfarek:
Hi,

This commit implements CRAM reading and writing in the `bam` module. The changes mostly relate to switching from the Reader and Writer structs using a `htslib::BGZF` struct to using a `htslib::htsFile` struct for BAM/CRAM file handling and updating the necessary htslib function calls. The reference can be set on a Reader or Writer struct by calling `set_reference()` with the path to the FASTA reference file.

Reading CRAM: Reading a CRAM file is mostly the same as reading a BAM file, with the addition of a call to `set_reference()` after creating a new `Reader` or `IndexedReader` and before attempting read operations.

Writing CRAM: A CRAM file can be opened for writing with `Writer::from_cram_path()` and the reference can be set with `set_reference()` before write operations.

With these changes I have been able to use `Reader`, `IndexedReader`, and `Writer` to read and write CRAM files and perform some other operations on them. You probably want to test these changes yourself to ensure I haven't missed anything obvious. All of the existing cargo tests pass when I run them. I imagine new tests for CRAM functionality would need be written up for this as well.


johanneskoester:
Thanks a lot, this is really appreciated. May I ask you to run cargo fmt? Also would it be possible for you to add a simple test case with a small CRAM file?

I am not so sure about the API now. Is `set_reference` always needed, or can it also be taken from the CRAM header? What do you think about making `set_reference` `pub(crate)`, and creating a cram module with  Reader, Writer IndexedReader wrapper types, that just delegate to bam, but require to set the reference as part of the constructor methods?

jfarek:
I am open to your suggestions for how the API should be structured. I structured these changes this way mainly because the htslib htsFile functions are mostly transparent between BAM and CRAM and to preserve the existing BAM tests here. After some testing, it appears that `set_reference` is indeed optional; it would only fail trying to read a CRAM if a reference path was not set and if the reference URL in the header was invalid.

I agree that this would be better implemented as a separate cram module. I am somewhat new to Rust but I'll see if I can write up a module for this as you described. In the meantime I'll add some CRAM tests to the bam module in this branch.

johanneskoester:
Wait, if set_reference is indeed optional, I think the API it can stay like it is now. The only issue would be naming, but I think that is fine, given that bam is the most abundant for now.

johanneskoester:
After this is merged, I might rename the readers and writers to `AlignmentFile` and `VariantFile`, similar to pysam. This would better reflect that they are in fact independent of the underlying variant of the format. However, let's do that at a later stage.

johanneskoester:
Or we have wrapper types for each format variant (VCF, BCF), (BAM,SAM,CRAM). I'll try to figure out what is better.

jfarek:
I implemented two tests here for CRAM reading and writing.

juliangehring:
@jfarek Does htslib offer some functionality to generate the cram header (especially the M5 checksums) from the reference file? This is probably far beyond the scope of your PR, just asking out of curiosity.

jfarek:
@juliangehring I think calling `sam_hdr_write()` handles filling in the UR and M5 tags in an htsFile header when converting from BAM or SAM to CRAM. I haven't poked around the CRAM internals enough to know if filling in the UR and M5 tags is something that can be called directly on a CRAM header.

juliangehring:
@jfarek Nice, thanks for the info!

jfarek:
Note for future development: htslib recently accepted a PR for an issue I ran into in implementing these tests (PR [here](https://github.com/samtools/htslib/pull/844)), involving how CRAM decodes NM tag integer type. The test BAM I put here will need to be regenerated from the test CRAM for these tests to continue to work with this htslib change, e.g. `samtools view -T test_cram.fa -b -o test_cram.bam test_cram.cram` with updated samtools+htslib.

PR #125, opened 2019-04-15
fix the handling of cstring pointers for sam/bam/cram index building

dlaehnemann:
* For pointers to a CString to work, the (unwrapped) CString needs
  to explicitly be in a variable in the scope of the pointer
  reference. So sorry for the funny let and if/else structure,
  there doesn't seem to be a cleaner syntax to do this.
* To test index creation with the default file name, I created
  a copy of the default test.bam to avoid accidentally overwriting
  its existing index.
* cargo fmt  ( rustfmt 1.0.3-stable (d6829d6 2019-02-14) )


dlaehnemann:
I think @johanneskoester @christopher-schroeder have the powers to merge, here. The sooner this is merged, the sooner #124 can move further... :)

PR #129, opened 2019-04-25
implement sequence-read trait

johanneskoester:


johanneskoester:
@dlaehnemann are the failing tests here related to what you fixed recently? I don't recall exactly what was the problem at that time.

dlaehnemann:
Jep, it's exactly the same error as before. And looking at the index code, that's the old one. So it looks like you need to rebase your branch onto the current master (the network graph confirms this: https://github.com/rust-bio/rust-htslib/network).

johanneskoester:
Thanks. Weird, I considered that, but then I did not see an update branch button here, which made me assume that it is in sync with the master branch... Anyway, let's see if it works now.

FelixMoelder:
Just tested the trait on rbt [consensus_from_bam](https://github.com/rust-bio/rust-bio-tools/commits/consensus_from_bam)-branch and everything looks fine now.

Issue #128, opened 2019-04-17
Beyond wrapping htslib


TyberiusPrime:
Dear authors,

pysam greatly enhances htslib with functions that for example find all pairs of aligned positions between a read and the reference, or all aligned blocks, or the splice sites defined by a read.

I've implemented some of them for my own usage and was wondering
whether such a contribution would be welcome, or if you'd rather have them live in a separate module.

johanneskoester:
That would be very welcome! Thank you!

PR #131, opened 2019-05-10
Cigar interpretation

TyberiusPrime:
This introduces cigar-interpreting methods inspired by pysam, as suggested in #128 

I have implemented it as an extension trait on the Record, as to separate the htslib and extension functionality cleanly.

I believe I have covered pretty much all the cigar related functionality pysam introduced.
The interface is similar, though I've rustified it - some of the Python methods  return None based on boolean parameters, those I split into two methods to avoid having the user deal with a bunch of Options if she doesn't have to.

My apologies for the test cases being so extraordinarily long - they're basically data driven and derived with pysam. Could refactor them into data files, but I don't see the point.

As always, I'm new to rust and do appreciate feedback!

Thank you for considering another puppy to include in rust-htslib.


TyberiusPrime:
Sure thing. I had tried to stay in line with pysam, but you're right, no need for that (and it's in the doc strings for those of us coming from python).

PR #132, opened 2019-05-15
validate fileformat before creating bam::Reader

### LINKED ISSUE #130: Handle invalid files in the BAM reader
The `bam::Reader` currently crashes when the input is not a BAM/CRAM/SAM file. For example, if one accidentally tries to read a VCF file:

```rust
Reader::from_path("test/test_various.vcf")
```

the entire program aborts fairly ungracefully. Is there a nice way of catching this, and e.g. returning a variant of a `ReaderPathError`? I guess one could manually check the first few bytes of the file before creating a Reader, but does htslib have some built-in capabilities for this that one could leverage?

TyberiusPrime:
This is supposed to fix #130.

We're using the fact that hts_open already calls hts_detect_format and are just checking that the format it detected is one of those we want to parse at this point.

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -155,6 +155,7 @@ impl Reader {
     /// * `path` - the path to open. Use "-" for stdin.
     fn new(path: &[u8]) -> Result<Self, BGZFError> {
         let htsfile = try!(hts_open(&ffi::CString::new(path).unwrap(), b"r"));
+
         let header = unsafe { htslib::sam_hdr_read(htsfile) };
         Ok(Reader {
             htsfile: htsfile,
@@ -778,6 +779,13 @@ fn hts_open(path: &ffi::CStr, mode: &[u8]) -> Result<*mut htslib::htsFile, BGZFE
     if ret.is_null() {
         Err(BGZFError::Some)
     } else {
+        if !mode.contains(&b'w') {
+            unsafe {
+                if (*ret).format.category != htslib::htsFormatCategory_sequence_data {
+                    return Err(BGZFError::Some);
+                }
+            }
+        }
         Ok(ret)
     }
 }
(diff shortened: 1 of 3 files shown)


src/bam/mod.rs
```diff
@@ -155,6 +155,15 @@ impl Reader {
     /// * `path` - the path to open. Use "-" for stdin.
     fn new(path: &[u8]) -> Result<Self, BGZFError> {
         let htsfile = try!(hts_open(&ffi::CString::new(path).unwrap(), b"r"));
+        unsafe {
+            if (*htsfile).format.format != htslib::htsExactFormat_sam
```
dlaehnemann: From a quick glance at [hts_detect_format()](https://github.com/samtools/htslib/blob/5c7ff9c6812f1e6ef1fd83ba73b8e42cfbabba93/hts.c#L199), it looks to me like you could substitute the three checks for `format.format` here with one check for `format.category`, i.e.:
```
if (*htsfile).format.category != htslib::htsFormatCategory_sequence_data
```
This should be more general and hopefully more future-proof, should htslib e.g. invent a newer sequence_data format...

Or am I overlooking something, here?

TyberiusPrime: No you're right.

TyberiusPrime:
> Some nice digging into htslib! ;)
> 
> I was also wondering, if we shouldn't include the corresponding check for `sam`, `bcf` and `tbx`, while you are at it?

Yes on bcf and tbx, for sam we only have a writer - no check necessary .

tbx: The format depends on the input file - some of the existing bed files are identified as sequence_data/sam actually,' while others identify as region list. 
I've implemented it to accept both.

Bonus bug: tbx crashed on non-existing files. That's fixed as well.

juliangehring:
This looks very nice and good to merge!

TyberiusPrime:
Thanks for the quick acceptance, you guys are cool.

PR #133, opened 2019-05-19
Fix `rid2name` out-of-bound errors in bcf module

juliangehring:
`bcf::header::rid2name(rid)` leads to segfaults if `rid` is larger than the number of contigs defined in the VCF header (`htslib::BCF_DT_CTG`). This is caused by an out-of-bound array access within an unsafe block.

The core part of the fix checks if the requested `rid` is within the range of defined contigs, and returns an error otherwise. This required changing `rid2name` to return an `Result<u32, RidIndexError>` since the previous return type (`u32`) didn't allow signalling the failure case. Please note that is a breaking change.

Summary of changes:
- Checks for valid `rid` in `rid2name`: This prevents the segfaults due to the out-of-bound access.
- Changes the return type of `rid2name` from `u32` to `Result<u32, RidIndexError>`, analogous to its counterpart `name2rid`. This is a breaking change!
- Introduces a new error variant `RidIndexError:: UnknownIndex(u32)`, analogous to the existing `RidError: UnknownSequence`.
- Adds a new public `contig_count` method which returns the number of defined contigs, similar to `sample_count`. This can be handy for checking the number of contigs or iterating over all contig names together with `rid2name`.
- Adds test cases for valid and invalid queries for `rid2name` and `name2rid`.

### DIFF
--- a/src/bcf/header.rs
+++ b/src/bcf/header.rs
@@ -240,6 +240,7 @@ impl HeaderView {
         unsafe { (*self.inner) }
     }
 
+    /// Get the number of samples defined in the header.
     pub fn sample_count(&self) -> u32 {
         self.inner().n[htslib::BCF_DT_SAMPLE as usize] as u32
     }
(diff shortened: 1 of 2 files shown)


src/bcf/header.rs
```diff
@@ -481,6 +491,10 @@ quick_error! {
             description("unknown sequence")
             display("sequence {} not found in header", name)
         }
+        UnknownRID(rid: u32) {
```
johanneskoester: Can you make this a separate error? The reason is that neither rid2name nor name2rid can throw both errors.

juliangehring: Yes, that makes sense. It's now has a dedicated `RidIndexError::UnknownIndex` error.

PR #137, opened 2019-05-28
Store BCF field size in `TagLength::Fixed`

juliangehring:
This PR extends the `TagLength::Fixed` enum to store the fixed number of fields defined in the B/VCF header. E.g. the size of the info header line ` INFO=<ID=I16,Number=16,Type=Float>` is now accessible as `TagLength::Fixed(16)`. I have found this very useful when parsing and validating VCF records. Getting the fixed field size comes with no additional compute costs, and fits well into the `TagLength` enum (note that the field size is not constant for variants other than `Fixed`). This also adds a number of test cases for tag type and length, based on the existing test files.

### DIFF
--- a/src/bcf/header.rs
+++ b/src/bcf/header.rs
@@ -296,7 +296,7 @@ impl HeaderView {
         tag: &[u8],
         hdr_type: ::libc::c_uint,
     ) -> Result<(TagType, TagLength), TagTypeError> {
-        let (_type, length) = unsafe {
+        let (_type, length, num_values) = unsafe {
             let id = htslib::bcf_hdr_id2int(
                 self.inner,
                 htslib::BCF_DT_ID as i32,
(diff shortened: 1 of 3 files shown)


src/bcf/header.rs
```diff
@@ -296,7 +296,7 @@ impl HeaderView {
         tag: &[u8],
         hdr_type: ::libc::c_uint,
     ) -> Result<(TagType, TagLength), TagTypeError> {
-        let (_type, length) = unsafe {
```
juliangehring: Out of curiosity, what is the benefit of `_` for the `type` variable in this context?

johanneskoester: I think it is unused, and hence we silence the warning of allocating an unused variable here.

johanneskoester:
Thank you!

src/bcf/header.rs
juliangehring: It seems to be used since it is returned by the function, isn't it?

juliangehring: Ah, `type` is of course a keyword in rust - that explains it.

PR #139, opened 2019-06-01
Use selective caching in travis builds

juliangehring:
Caches the cargo binaries in travis builds which avoids recompiling cargo-travis and friends in every travis run. This saves ~10min in each run while keeping the cache small (<20MB).
See also:
- https://levans.fr/rust_travis_cache.html
- https://github.com/roblabla/cargo-travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,12 @@ language: rust
 rust:
   - stable
 
-sudo: false
+cache:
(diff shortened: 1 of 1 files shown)


juliangehring:
The decrease in coverage seems to be an artefact and unrelated to the PR.

Issue #142, opened 2019-06-18
Writing uncompressed BAM


wwood:
Hi,

Just wondering first of all if this is possible given the current API? My use case currently spends >50% of its time needlessly compressing records just to pipe them into a `samtools sort`. I couldn't see any possible way given the current API, but seems possible with the bindings.

Without bothering to implement anything, maybe add a `set_compression_level(i16)` method to `Writer` which modifies the internal htsFile -> htsFormat -> compression_level state after new is called by from_path or from_stdout? I've not worked with htslib binding before so that's just a guess.

Ta, ben

juliangehring:
@wwood I have been working on exactly the feature you are describing for BAM/CRAMs and BCF/VCFs. Most parts are in place, I just didn't find the time to finalise it and open a proper pull request. If you are not too much in a hurry, I should be able to open a pull request throughout the week - would also be happy to get your feedback on the API.

wwood:
Sure, sounds good to me. Great.

juliangehring:
I would focus on setting the level of the bgzip compression here. CRAM allows for a much wider range of compression codecs and options, but would require a lot of work.

In terms of the interface for setting the compression level, three options came to my mind:

1. Specifying the compression levels 0 to 9 as a plain integer - simple and straight forward. Essentially the same as e.g. htslib and samtools use, where 0 is uncompressed and 9 the maximum compression.
2. Specifying the compression levels -1 to 9 as a plain integer similar to above, where -1 encodes the default compression level (6). This is consistent with the internal representation in htslib. I'm not sure if there is a big need for specifying the default level explicitly, but maybe someone has a use for this? It feels however a bit unintuitive to me since one always has to explain what -1 represents and could get confused with the `-1` option in samtools.
3. Offer a higher-level interface that covers common user choices, together with an interface for specifying the raw levels for advanced users. Users might prefer options, such as "uncompressed", "high speed", "maximum compression" , over the explicit zlib compression levels. One could nicely represent both in an enum:
```rust
enum CompressionLevel {
    Uncompressed,  // 0: htslib::Z_NO_COMPRESSION
    Fastest, // 1: htslib::Z_BEST_SPEED 
    Standard, // 6: htslib::Z_DEFAULT_COMPRESSION
    Maximum, // 9: htslib::Z_BEST_COMPRESSION
    Level(i32), // specify your own level, for finer control and compatibility with htslib
}
```

[...]

PR #144, opened 2019-06-22
Support setting the compression level for `bam::Writer`

juliangehring:
Allow setting the compression level when writing BAM/CRAM files. Adds
- `Writer.set_compression_level()`
- `CompressionLevel` enum to specify the desired compression level
- `CompressionLevelError ` error enum
- test cases


juliangehring:
Should resolve #142.
@wwood, what do you think?

wwood:
Sure, looks good, for what my opinion is worth. This is enough for my use case, though I thought you were planning bcf too?

PR #145, opened 2019-06-22
Drop unused bcf::GT_MISSING constant

juliangehring:
Hardcoding the `GT_MISSING` constant in the `bcf` module doesn't seem to be needed any longer with newer htslib versions and, if needed, `htslib::bcf_gt_missing` can be used instead.


juliangehring:
@johanneskoester What exactly do you want to cover in the documentation? `GenotypeAllele` already seems to represent missing genotypes, but maybe I'm missing something.

johanneskoester:
You are right, we hide that detail already. I forgot about it.

PR #146, opened 2019-06-22
Code simplification and harmonisation

juliangehring:
A number of small code simplifications that should improve the readability and consistency. Most of them come from clippy warnings. All suggested changes are non-breaking - changes that would affect the user interface have been omitted here.


juliangehring:
@johanneskoester This is now ready for review.

PR #148, opened 2019-07-14
check for OOM during realloc & panic, rather than proceeding with null

pmarks:
Will avoid segfault and be easier to diagnose issues if you run into OOM.

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -388,7 +388,6 @@ impl Record {
     }
 
     fn realloc_var_data(&mut self, new_len: usize) {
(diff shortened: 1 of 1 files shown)


johanneskoester:
It just needs a `cargo fmt` now.

pmarks:
@johanneskoester fmt is fixed.  I don't think we can really test this codepath in CI unfortunately.

pmarks:
@johanneskoester - ok to merge this?

johanneskoester:
Thanks!

PR #152, opened 2019-08-06
Add wrapper for bam_itr_querys

ifiddes-10x:
Construct a new function for IndexedReader that wraps `bam_itr_querys` and allows for direct string queries. This allows for access to unmapped reads.

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -367,7 +367,8 @@ impl IndexedReader {
         if let Some(itr) = self.itr {
             unsafe { htslib::hts_itr_destroy(itr) }
         }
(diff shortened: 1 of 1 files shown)


src/bam/mod.rs
```diff
@@ -363,6 +363,21 @@ impl IndexedReader {
         }
     }
 
+    pub fn fetch_str(&mut self, region: ffi::CString) -> Result<(), FetchError> {
```
johanneskoester: I suggest to expect region as a Rust byte slice, i.e. `b"chr1:1-1000"`.

johanneskoester:
Great, thank you!

PR #153, opened 2019-08-07
Avoid references when copying is more efficient

juliangehring:
A number of small code simplifications which avoid unnecessary references, unneeded imports and add a few test cases. Please note that some of these are breaking changes.

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -14,7 +14,7 @@ use std::u32;
 
 use regex::Regex;
 
-use crate::bam::{AuxWriteError, HeaderView, ReadError};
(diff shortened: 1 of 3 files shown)


src/bam/record.rs
```diff
@@ -769,8 +769,8 @@ pub enum Cigar {
 }
 
 impl Cigar {
-    fn encode(&self) -> u32 {
-        match *self {
+    fn encode(self) -> u32 {
```
johanneskoester: Interesting! Are you sure this is idiomatic? Semantically, I would think a borrow makes more sense here.

PR #154, opened 2019-08-27
Reduce dependencies

juliangehring:
This reduces the number of dependencies which aren't really needed:
- ~~Removes `itertools` and replace it with `collect` in `std`~~
- Build `bindgen` with minimal features: The CLI of `bindgen` isn't needed for building
- Update dependencies with significant changes

Overall, the changes reduce the number of dependent crates to build from 109 to 92.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,5 +43,5 @@ pretty_assertions = "0.5.1"
 
 [build-dependencies]
 fs-utils = "1.0"
-bindgen = ">=0.37"
+bindgen = {version = "0.51.0", default-features = false}
 cc = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,11 +21,11 @@ itertools = "0.8"
 quick-error = "1.2"
 newtype_derive = "0.1"
 custom_derive = "0.1"
-url = "1.4"
+url = "2.1.0"
 ieee754 = "0.2"
-lazy_static = "0.2"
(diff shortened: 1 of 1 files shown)


juliangehring:
I don't understand why the coverage is dropping here: The files with decreased coverage don't have any changes. Probably an artefact again, similar to some previous PRs.

johanneskoester:
Thanks a lot. I am not sure whether removing itertools is sustainable though. Sure, now it is not needed, but it has a lot of convenient helpers that will becomes more and more important as we add sugar to the library. Hence, I would vote for keeping it. The rest is great.

juliangehring:
It seems easy to add `itertools` back in if it is needed. At the moment, `itertools` doesn't offer any benefits in terms of performance or ergonomics over using the standard library. I would suggest to take it out now until it is actually needed. Let me know what you think. If you think it should really stay in, would we only keep the dependency in `Cargo.toml` or also keep the `collect_vec()` calls?

johanneskoester:
To me, collect_vec seems a little easier on the eye than the type annotation in combination with collect. So I see a (minor though) benefit of keeping itertools. Since it is likely needed in the future, this outweights the removal benefit for me. So, I would say keep collect_vec() as it was before.

juliangehring:
`itertools` is back in. The coverage report still fails for the same reason as before; again an artefact.

johanneskoester:
Thanks a lot!

PR #163, opened 2019-10-31
rfc: Better iteration

TyberiusPrime:
# tl/dr: 
 * Rc based Iterators are on par with read(), have a nicer API. 
 * Records is 10% slower than read().
 * fetch_iter() is a more ergonomic than fetch/fetch_str.

# Prelude
This is an RFC - the code needs executive decisions, polishing, and the commits have yet to be cleaned up.

# Iterators
The recent extension of read() to error handling had me thinking:

## Read() vs records
I don't actually know how much the speed difference between records() and read() is.

The answer is: about 10%. (All benchmarks were on a BAM with 50k reads all on one reference (to avoid seeks within the file)).

## 0.25 api changes
```while let Ok(_) = bam.read()``` was bad, and 
```while let Ok(true) = bam.read()``` is a horrible, error eating anti pattern.

I researched and found that streaming iterators - ie. those returning values borrowed from self is difficult in Rust right now. 

From my point of view there are three approaches:
 1. abandon the Iter trait - e.g. rust-htslib read() method, the[ streaming-iterator  crate](https://crates.io/crates/streaming-iterator)
 2. Do [funky stuff with types](http://lukaskalbertodt.github.io/2018/08/03/solving-the-generalized-streaming-iterator-problem-without-gats.html) (just a tad bit beyond my current understanding of Rust )
3. have an Iterator that returns Result<Rc<Record>> and push the borrow checking into the runtime (see [reddit](https://www.reddit.com/r/rust/comments/d4rl3d/a_remarkably_simple_solution_to_the_problem_of/)

I've implemented 1. here as stream(), using StreamingIterator. It is essentially as fast as read(). The API isn't cleaner, though you gain the iterator combinators defined on StreamingIterator. Also the streaming-iterator crate doesn't seem to be super active.
(Note: The implementation here doesn't do error checking - I wanted to move on).

I've also implemented 3. as .rc_records().
It is only about **1%** slower than read()/stream()
(provided you don't force allocations by holding on to the Rcs). 
I guess Rust's Rc is very fast.

# fetch/fetch_str + read/records() 
I absolutely see where the dichotomy of fetching first, deciding how to iterate then came from.
There are two issues here - the first is that an enum with good From<> implementations can make 
fetch() much nicer to use, and the second one is that the Rc-Iterator is proably 'fast enough'
to just be the default.

I've implemented both as fetch_iter(Into<FetchDefinition>) -> RcRecords.
This allows constructs like
```for read in bam.fetch_iter(("chr1", 0, 1000)) {...}```
```for read in bam.fetch_iter("chr1") {...}``` (either &str or &[u8], full reference)
```for read in bam.fetch_iter("chr1:0-1000") {...}``` (auto detect reference name, if not pass to htslib)
```for read in bam.fetch_iter((tid: u32, 0, 1000)) {...}```
```for read in bam.fetch_iter(tid)```

And I believe that's much nicer than having to retrieve the tid, and the length from the header by yourself.

Looking forward to your feedback!

.

### CODE REVIEW

**Comment by johanneskoester** [comment]
Mhm, are you sure the current form of the `read()` method is an anti-pattern? It is very similar to the Read trait from the standard lib, just that it not returns the bytes but a flag whether something was read or not.

**Comment by johanneskoester** [comment]
Do you have an idea what makes the current `records()` 10% slower that `read()`. I always thought it would be the additional allocation of a new record for each item, but that does not seem to be the case since one needs to do the same for an Rc based iterator. Is it an additional copy that it needs when passing ownership to the caller?

**Comment by TyberiusPrime** [comment]
@johanneskoester  the beauty of the RC based iterator implemented here is that it does not allocate if the user didn't keep a reference! RC::get_mut() will return Some only if the reference count is 1 - ie. no other piece of code has a reference, and we can reuse it.

The `read()` method is fine, it's the usage with ```while let Ok(true) = bam.read() ```.
It will silently exit the loop on no more results (good) or an error (not good).
I'm not even sure where I picked this up (is it in the docs?).

You need to use 
```
loop { 
   match bam.read() {
      Ok(true) => {...},
      Ok(false) => break;
      Err(e) => panic!(e)
    }
}
```
which is kinda tedious.
Compare with 
```
for read in bam.rc_records().map(|r| r.expect("BAM failure"))
{
... 
}
```
(The same could be achieved by turning read() into a 'Streaming Iterator' from the crate of the same name. But it has way less cool iterator methods than the real thing).

**Comment by johanneskoester** [comment]
Actually, the non-iterator version is much simpler to use than your example, assuming that you are inside a function that returns a result:
```
if !bam.read(&mut record)? {
      return Ok(());
}
```

However, this does not mean that an iterator wouldn't be a welcome addition. I just have to dig into your PR to understand why it is advantageous to use Rc here.

TyberiusPrime:
Can we possibly move forward on the enum + From<> for fetch independent of the RC stuff?

johanneskoester:
Yes, the enum+From approach for fetch is great. I am happy to take that as a separate PR first.

brainstorm:
I took the liberty to fix the changes respective to `master` before this PR diverges too much, hope you don't mind?

brainstorm:
Thanks @dlaehnemann and @TyberiusPrime I was just scanning the backlog of PRs and trying to bump them up somehow, there's very good code here that would be a pity to see rot if left unattended for too long ;)

TyberiusPrime:
Closing this in favor of #265 and #261.

PR #164, opened 2019-11-01
fixing clippy lints (+breaking change)

TyberiusPrime:
This makes clippy useful for rust-htslib,
by making it shut up about the bindgen generated code,
and fixing all the minor suggestions it had.

The only visible change is that CigarString::from_bytes
and CigarString::from_str become TryFrom trait implementations.
Under todays rust, they were arguably misnamed (idiomatic frotm mustn't fail)
and should've been From/TryFrom implementations.
This means that users need to convert from CigarString::from_bytes(x) to
CigarString::try_from(x), but the functionality is untouched.

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -31,10 +31,10 @@ pub use crate::bam::header::Header;
 pub use crate::bam::record::Record;
 
 /// Implementation for `Read::set_threads` and `Writer::set_threads`.
-pub fn set_threads(htsfile: *mut htslib::htsFile, n_threads: usize) -> Result<()> {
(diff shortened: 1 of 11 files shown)


src/bam/mod.rs
```diff
@@ -31,18 +31,18 @@ pub use crate::bam::header::Header;
 pub use crate::bam::record::Record;
 
 /// Implementation for `Read::set_threads` and `Writer::set_threads`.
-pub fn set_threads(htsfile: *mut htslib::htsFile, n_threads: usize) -> Result<()> {
+pub unsafe fn set_threads(htsfile: *mut htslib::htsFile, n_threads: usize) -> Result<()> {
```
johanneskoester: Did clippy suggest to move the unsafe here? I don't understand the rationale behind that. Do you have a clue?

PR #165, opened 2019-11-06
Add basic functionality to read indexed FASTA files

landesfeind:
Hi,
this pull request adds basic FAIDX functionality. Importantly, not all functions are supported/wrapped but enough for my use case. 

I tried to follow your coding conventions but please don't hesitate to point out anything which can be improved.

Best,
Manuel


src/faidx/errors.rs
```diff
@@ -0,0 +1,13 @@
...
+use std::path::PathBuf;
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[derive(Snafu, Debug, PartialEq)]
+#[snafu(visibility = "pub")]
+pub enum Error {
+    #[snafu(display("file not found: {}", path.display()))]
+    FileNotFound { path: PathBuf },
+    #[snafu(display("invalid (non-unique) characters in path"))]
```
dlaehnemann, line 11: Did you mean `non-unicode`, here?

src/faidx/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Create a new Reader from path.
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - the path to open.
+    pub fn from_path<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
+        Self::new(&path_as_bytes(path, true)?)
+    }
+
+    /// Create a new Reader from URL.
```
dlaehnemann, line 49: It's fairly obvious, but maybe add the `# Arguments` section to the Doc string?

src/faidx/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+mod tests {
+	use super::*;
+
+	fn open_reader() -> Reader {
+		Reader::from_path(format!("{}/test/test_cram.fa", env!("CARGO_MANIFEST_DIR"))).ok().unwrap()
+	}
+
+	#[test]
+	fn faidx_open() {
+	}
```
dlaehnemann, line 101: This is probably a leftover or was intended as an additional test?

src/faidx/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Create a new Reader.
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - the path to open
+    fn new(path: &[u8]) -> Result<Self, Error> {
+        let cpath = ffi::CString::new(path).unwrap();
+        let inner = unsafe { htslib::fai_load(cpath.as_ptr()) };
+        Ok(Self { inner })
+    }
```
dlaehnemann, line 63: What is the distinction between this `new()` function to create a `Reader` and the `from_path()` function above? From what I can see, only the `from_path()` above is being tested below. So maybe this is an earlier attempt? Or does it provide another benefit and should also be tested?

src/faidx/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+            .to_str()
+            .ok_or(Error::NonUnicodePath)?
+            .as_bytes()
+            .to_owned())
+    } else {
+        Err(Error::FileNotFound {
+            path: path.as_ref().to_owned(),
+        })
+    }
+}
```
dlaehnemann, line 32: If this function might be useful for other `Reader`s, it could go into `utils.rs`?

src/faidx/mod.rs
landesfeind: I agree. But I followed the general pattern of the lib. For example, there is also `bam/errors.rs` etc. If all maintainer agree, then we can start combining this.

src/faidx/mod.rs
landesfeind: I agree. But I followed the general pattern of the lib. For example, there is also `bam/errors.rs` etc. If all maintainer agree, then we can start combining this.

landesfeind:
Dear @dlaehnemann, thank you very much for the review! 
Since this pull request was generated quite a while ago and a lot of changes were introduced in the master, I cherry picked this pull-request into a separate branch and generate a new pull request. This makes further improvement easier. The superseding pull request is #214

src/faidx/mod.rs
dlaehnemann: I think this still remains in the new branch?

src/faidx/mod.rs
dlaehnemann: Jep, I guess generic errors such as `FileNotFound` and `NonUnicodePath` could be maybe be moved up to a new `src/errors.rs` to avoid redundancies.

And `path_as_bytes` could easily live alongside `path_to_cstring` in `utils.rs`:
https://github.com/rust-bio/rust-htslib/blob/61b0644b73ab40612cd23e104ff0a3fe647bf064/src/utils.rs#L30-L34

src/faidx/mod.rs
landesfeind: Yes it does but now does test the (test-)file opening. Should be the first test because if it fails, e.g., if the test-file is missing, all subsequent tests will fail too. Hope this will make debugging simpler - test simple things first :)

src/faidx/mod.rs
landesfeind: Yes, I can do that but its a larger work-package (considering a harmonization with the other `**/errors.rs`) and maybe we want to separate this from the faidx implementation? How about merging as-is and creating a separate issue/enhancement with an according branch to work on the generalization of errors. If you are willing to review more pull-request from me, I am open to work on the new errors-issue :-)

src/faidx/mod.rs
dlaehnemann: Ah, yes, I missed that little difference. :D

src/faidx/mod.rs
dlaehnemann: Jep, sounds like a good way forward. I create issue #215 for putting general errors in a new `src/errors.rs`. And I guess having that is then a prerequisite for putting `path_as_bytes` in `src/utils.rs`, as that is using exactly those errors. So I've added that into the new issue, as well. And then, let's move forward with #214.

PR #166, opened 2019-11-11
Add basic example for read method

thomasmulvaney:


pmarks:
Great, thanks for the contribution. The coveralls regression is a false positive. Merging.

PR #170, opened 2019-12-02
Add method to retrieve all tags of a BAM record

jch-13:
This PR allows getting all tags from a bam::Record as a Vec of tuples. I am not sure this implementation is the right way to do it, so please be sceptical :)

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -601,6 +601,42 @@ impl Record {
         }
     }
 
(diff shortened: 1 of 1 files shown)


jch-13:
Sure, good idea! While writing tests covering e.g. `Aux::Integer(i64::MAX)` I discovered #209, which is probably best resolved before proceeding here...

jch-13:
Closed in favor of #292

PR #172, opened 2020-01-15
bam::Record: make serde faster

nlhepler:


nlhepler:
I fix the formatting separately in #173

PR #177, opened 2020-01-17
inline bam1_t into bam::Record

nlhepler:
This may be a bit controversial, but because (on 64-bit platforms) `bam1_t` is just 52 bytes (54 if not BAM_NO_ID), it doesn't make much sense to incur the pointer indirection/allocation/fragmentation all the time rather than just keeping them around on the stack.

I will work on some benchmarks to give this greater context, but I did want to put it here for greater discussion.

It does mean the `from_inner(inner: *mut htslib::bam1_t)` constructor is a bit weird now--it copies the data over but keeps using the old `data` ptr until the next re-alloc, when `own` becomes `true`.

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -6,6 +6,7 @@
 use std::convert::TryFrom;
 use std::ffi;
 use std::fmt;
+use std::mem::{size_of, MaybeUninit};
 use std::ops;
 use std::slice;
 use std::str;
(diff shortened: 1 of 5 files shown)


nlhepler:
This also contains a fix for a mutability bug: `remove_aux(&self)` is now `remove_aux(&mut self)`-- the risks of default interior mutability.

src/bam/record.rs
```diff
@@ -136,7 +150,11 @@ impl Record {
         };
 
         let succ = unsafe {
-            htslib::sam_parse1(&mut sam_string, header_view.inner_ptr_mut(), record.inner)
+            htslib::sam_parse1(
+                &mut sam_string,
+                header_view.inner_ptr_mut(),
+                &mut record.inner as *mut htslib::bam1_t,
```
johanneskoester: For all these pointer conversions, can't we have a helper method?

nlhepler: Good point, updated.

PR #181, opened 2020-01-31
Lh/leaked hdr mutability

nlhepler:
htslib::sam_parse1 may modify the header, so this was leaking interior mutability. Not terribly high-risk, but still not safe.

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -92,6 +92,9 @@ pub trait Read: Sized {
     /// Return the header.
     fn header(&self) -> &HeaderView;
 
+    /// Return the header, mutable.
+    fn header_mut(&mut self) -> &mut HeaderView;
+
     /// Seek to the given virtual offset in the file
     fn seek(&mut self, offset: i64) -> Result<()> {
         let htsfile = unsafe { self.htsfile().as_ref() }.expect("bug: null pointer to htsFile");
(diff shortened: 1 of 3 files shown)


nlhepler:
I thought about that, but I the only use I saw was in the tests, but you're probably right.

nlhepler:
OK, I added the mutable accessor.

There was some other weird stuff going on, I addressed that also.

src/bam/mod.rs
```diff
@@ -811,8 +822,8 @@ impl HeaderView {
     }
 
     #[inline]
-    pub fn inner(&self) -> htslib::bam_hdr_t {
-        unsafe { (*self.inner) }
+    pub fn inner(&self) -> &htslib::bam_hdr_t {
+        unsafe { self.inner.as_ref().unwrap() }
     }
```
nlhepler: Like, this was weird. `inner()` wasn't returning a reference to the interior, but a copy of the object.

Some places would access a member of the copy, still others would pass that copy as a mutable reference into htslib functions.

src/bam/mod.rs
```diff
@@ -420,10 +427,10 @@ impl IndexedReader {
         data: *mut ::std::os::raw::c_void,
         record: *mut htslib::bam1_t,
     ) -> i32 {
-        let _self = unsafe { &*(data as *mut Self) };
```
nlhepler: This is also weird: these `pileup_read` C functions cast to mutable self, dereference (now we have a copy?) then take the reference of that.

I converted these to `as_ref().unwrap()` or `as_mut().unwrap()` as appropriate.

pmarks:
@nlhepler looks like there's a couple merge conflicts

PR #182, opened 2020-02-01
don't use default bindgen features -- cuts down on dependency list

pmarks:
removes a few dependencies, particularly `clap` which takes a while to compile.


dlaehnemann:
Sorry for not providing feedback here, for some reason I only got the request for review today. But I don't think there's anything speaking against cutting down on dependencies and the tests should catch any potential problems on this level. So it's good that you went ahead!

PR #185, opened 2020-02-04
when including hts-sys, don't use default features -- let the outer r…

pmarks:
let the outer rust-htslib crate control them.  Without this change, there's no way to turn off usage of lzma and bzip which aren't needed if only BAM support is needed.


pmarks:
@dlaehnemann good call - added more details about the available `features` to the README

PR #188, opened 2020-02-19
bam::Record::set: Preserve aux data after set

sjackman:


src/bam/record.rs
```diff
@@ -321,22 +320,29 @@ impl Record {
...
-            + cigar_width
-            + ((seq.len() as f32 / 2.0).ceil() as usize)
-            + qual.len()) as i32;
-
-        assert!(qname.len() <= 256);
-
+        let orig_aux_offset = self.qname_capacity()
+            + 4 * self.cigar_len()
+            + (self.seq_len() + 1) / 2
+            + self.seq_len();
```
nlhepler: OK, cool, just for the record, there are l_seq 0xFF bytes for QUAL when it is "unset".

src/bam/record.rs
```diff
@@ -321,22 +320,29 @@ impl Record {
         let q_len = qname.len() + 1;
         let extranul = extranul_from_qname(qname);
 
-        self.inner_mut().l_data = (q_len
-            + extranul
-            + cigar_width
-            + ((seq.len() as f32 / 2.0).ceil() as usize)
-            + qual.len()) as i32;
-
-        assert!(qname.len() <= 256);
```
nlhepler: we do probably still want this assert _somewhere_.

src/bam/record.rs
```diff
@@ -321,22 +320,29 @@ impl Record {
...
-
+        let orig_aux_offset = self.qname_capacity()
+            + 4 * self.cigar_len()
+            + (self.seq_len() + 1) / 2
+            + self.seq_len();
+        let new_aux_offset = q_len + extranul + cigar_width + (seq.len() + 1) / 2 + qual.len();
+        let aux_len = self.inner.l_data as usize - orig_aux_offset;
```
nlhepler: OK, I do believe that is is _impossible_ for `l_data` to be `< orig_aux_offset`. Subtractions of unsigned like this are always scary--probably best to communicate this in a comment.

src/bam/record.rs
sjackman: I've moved it up to new line 310. `assert!(qname.len() < 255);` The SAM spec says it has to be strictly less than 255.

src/bam/record.rs
sjackman: Agreed. I'll add an assert `assert!(orig_aux_offset <= self.inner.l_data);`

PR #189, opened 2020-02-22
Features & build cleanups

pmarks:
- Use `cc` and a `config.h` generated by `build.rs` to compile htslib
- Make all the optional features work correctly (on or off)
- Make bindgen optional by bundling pre-made bindings (huge reduction in clean build time)
- Add optional support for `libdeflate` and pipe that through to htslib.

TODO:
- [x] fix generation of version.h
- [x] switch to official `libdeflater` once it's updated with required changes. (PR at adamkewley/libdeflater#4)


brainstorm:
Thanks for this @pmarks! It seems that it is now failing (against master w/ htslib 1.10.x) because of `curl.h` et al:

https://github.com/rust-bio/rust-htslib/runs/729762497?check_suite_focus=true#step:9:114

I'll try to get my PR finished and passing (#193) ASAP so that this doesn't become a blocker to merge this PR.

pmarks:
@brainstorm I have `libdeflate` working correctly in this PR.

brainstorm:
Woah, great work @pmarks, checking it out right now...

hts-sys/build.rs
```diff
@@ -3,34 +3,80 @@
...
-            .unwrap()
-            .success()
-        {
-            panic!("failed to strip {} support", feature);
-        }
-    }
-}
+
+// these need to be kept in sync with the htslib Makefile
+const FILES: &[&str] = &[
```
brainstorm, line 17: I wonder if missing `hfile_s3.c` disables AWS S3 functionality? Personally I would really like to have it included in the mix, TBH.

hts-sys/build.rs
```diff
@@ -43,114 +89,115 @@ fn main() {
...
     }
 
-    let tool = cfg.get_compiler();
-    let (cc_path, cflags_env) = (tool.path(), tool.cflags_env());
-    let cc_cflags = cflags_env.to_string_lossy().replace("-O0", "");
-    let host = env::var("HOST").unwrap_or_default();
-    // autoreconf & ./configure (with no args) steps are necessary to include the htslib plugins (hfile_s3.o, hfile_s3_writer.o, etc...)
-    if  // cleanup first
-        // TODO: Have top level "cargo clean" do this instead of in here, see:
-        // https://github.com/rust-lang/cargo/issues/572#issuecomment-632456478
```
brainstorm, line 84: No need/wish to have `cargo clean`? I did affect my builds when re-compiling rust-htslib :-S

hts-sys/build.rs
brainstorm: Since we now have htslib 1.10.2 on `master`, it should be possible to do things like consult BAM headers sitting on S3 buckets directly like this:

https://gist.github.com/brainstorm/83ab0fe91b3aaa1e73072e58319d15bb#file-s3_htslib-c-L16

hts-sys/build.rs
```diff
@@ -43,114 +89,115 @@ fn main() {
...
-    fs::copy(out.join("htslib").join("libhts.a"), out.join("libhts.a")).unwrap();
 
     println!("cargo:root={}", out.display());
     println!("cargo:include={}", include.display());
     println!("cargo:libdir={}", out.display());
-    println!("cargo:rustc-link-lib=static=hts"); // XXX: only for static, adapt for dynamic?
     println!("cargo:rerun-if-changed=wrapper.c");
     println!("cargo:rerun-if-changed=wrapper.h");
-    println!("cargo:rerun-if-changed=htslib/Makefile");
-    let globs = std::iter::empty()
```
brainstorm, line 147: This essentially un-does PR https://github.com/rust-bio/rust-htslib/pull/204 from @nlhepler , but I guess it's fine with the new approach?

src/bam/mod.rs
```diff
@@ -1720,14 +1720,24 @@ CCCCCCCCCCCCCCCCCCC"[..],
...
                 }
                 fs::metadata(output_bam_path).unwrap().len()
             })
             .collect();
 
         // check that out BAM file sizes are in decreasing order, in line with the expected compression factor
+        println!("testing compression leves: {:?}", levels_to_test);
+        println!("got compressed sizes: {:?}", file_sizes);
+
+        // libdeflater comes out with a slightly bigger file on Max compression
```
brainstorm, line 1735: How much bigger if you have the numbers handy? Just curious :)

hts-sys/build.rs
pmarks: Added features for s3 and gcs.

hts-sys/build.rs
pmarks: The cargo bug is closed & it doesn't seem like it's going to be addressed.  Since we're moving the source into `OUT_DIR` and using `cc::Build.compile()` the object files and the copy of the source will actually get deleted by `cargo clean`, so I think we're good.

hts-sys/build.rs
pmarks: hmm, in theory we should still rebuild if a header changes? I'll ask him about that.

src/bam/mod.rs
pmarks: It's something like 450 bytes on Maximum and 438 bytes on Level 6 for this tiny example.

hts-sys/build.rs
pmarks: @brainstorm we should add some unit tests for s3 / gcs -- feel free to drop them into this PR, or we can do as a follow-up.

hts-sys/build.rs
nlhepler: We will want to keep the rerun-if-changed directives for header files, as those are effectively dumped into the C files that include them by the preprocessor.

pmarks: @nlhepler restored.

Cargo.toml
```diff
@@ -33,12 +33,15 @@ snafu = "0.6.8"
...
+default = ["bzip2", "lzma", "curl", "bindgen"]
 bzip2 = ["hts-sys/bzip2"]
 lzma = ["hts-sys/lzma"]
+bindgen = ["hts-sys/bindgen"]
 curl = ["hts-sys/curl"]
+libdeflate = ["hts-sys/libdeflate-sys"]
+s3 = ["hts-sys/s3"]
+gcs = ["hts-sys/gcs"]
 serde = ["serde_base", "serde_bytes"]
-static = []
```
brainstorm, line 41: @pmarks I just noticed this... all builds will be `static` by default now? I'm fine with since I need MUSL builds to be static but perhaps folks want to i.e link against a modified locally installed `zlib`, for instance?

hts-sys/build.rs
```diff
@@ -43,115 +90,164 @@ fn main() {
...
-        .arg("clean")
-        .status().unwrap().success()
+    let use_gcs = env::var("CARGO_FEATURE_GCS").is_ok();
+    if use_gcs {
+        config_lines.push("#define ENABLE_GCS 1");
+        cfg.file("htslib/hfile_gcs.c");
+        println!("cargo:rerun-if-changed=htslib/hfile_gcs.c");
+    }
+
+    let use_s3 = env::var("CARGO_FEATURE_S3").is_ok();
```
brainstorm, line 166: @pmarks @nlhepler To be honest, I would not use environment variables for configuration of features but cargo flags instead (as I did above with `static = []`), see https://github.com/alexcrichton/xz2-rs/pull/63 for reference.

hts-sys/build.rs
brainstorm: Agree, testing the S3 part of that, getting the following when trying to read a header from S3 though:

```
[E::hts_open_format] Failed to open file "s3://umccr-research-dev/test.bam" : Function not implemented
```

Working on it...

hts-sys/build.rs
brainstorm: IMO it's more cargo-intuitive (as a developer) to use `features=` instead of having to export something pre-compilation.

nlhepler: @brainstorm this is how cargo features are passed to the build script, see [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts)

nlhepler: See e.g. https://github.com/rust-lang/libc/blob/master/build.rs#L5-L19 for an example.

brainstorm: @nlhepler I hear you and I just thought about additional reasons to use simple cargo flags **in addition to env vars**. 

For instance, we have to define each one of them again under the cross passthrough if we want to have them available during builds and tests:

https://github.com/rust-bio/rust-htslib/blob/master/Cross.toml#L2

If Cargo flags are used (actually, in addition to env vars like in that `xz2-rs` crate I referred to above?), users could just have: **features=["s3", "gcs"]**, etc on their `Cargo.toml`, without the need to modify `Cross.toml` when and if new features are introduced in (rust-)htslib.

Cargo.toml
pmarks: agreed, we should retain. fixed.

johanneskoester:
@pmarks @brainstorm I trust your judgement here, as you are both way better in building C than me. Feel free to merge whenever you feel this is done.

hts-sys/build.rs
nlhepler: @brainstorm that's exactly the mechanism that I'm trying to explain here. Cargo builds the build.rs script, and when it runs it, ensures env::vars CARGO_FEATURE_S3 and CARGO_FEATURE_GCS are defined in the build script's execution environment if you passed `--features=s3,gcs` to `cargo b`. There is no mucking around with e.g. `CARGO_FEATURE_S3=1 cargo b --features=s3`. These env vars are the mechanism by which cargo communicates to the build script.

brainstorm: Ahaa, sorry @nlhepler, thanks for the clarification and patience ;)

nlhepler: :) My bad, I was doing a poor job of explaining. I think the advantage of the CARGO_FEATURE approach is that it doesn't require the build script to be rebuilt upon feature changes, just re-run.

brainstorm: Definitely, I'm sold now ;)

hts-sys/build.rs
brainstorm: Feel free to add in the `libdeflate-sys` new release and merge it, LGTM. Looks like the S3 stuff needs a bit more attention (for my use case at least).... can be done post-merging this, TBH... so, ship it if you want/can! ;)

pmarks:
OK @johanneskoester @brainstorm @nlhepler, this should be ready to go.

Last question what do you think about taking `bindgen` out of the default dependency list?  

In our experience compiling bindgen then rust-htslib become a huge part of the critical path of build time.  I think there's also some machines that don't have the required clang installation that would now work out of the box.  I'm not aware of any downsides.

brainstorm:
I'm very happy with shorter build times indeed... only thing I can think of is whether we will face any kind of operational/caching issue re-generating those prebuilt bindings in the future? I guess it should be mentioned somewhere in docs?

Slightly OT but relevant since the rebuilding of `build.rs`: I'm getting consistent `ENOSYS` (a.k.a `Function not implemented`) on `features="s3"` when just trying to read a BAM header from S3, so there's something off while building that `hfile_s3.c` AWS S3 support:

```
[E::hts_open_format] Failed to open file "s3://test-bucket/test.bam" : Function not implemented
```

Equivalent C code fine though:

```C
$ cat s3_htslib.c
#include <stdio.h>
#include <stdlib.h>

#include <htslib/hfile.h>
#include <htslib/hts.h>
#include <htslib/sam.h>

#include <string.h>
#include <unistd.h>

int main(int argc, char **argv)
{
	samFile *fp_in = hts_open(argv[1],"r"); //open bam file
	bam_hdr_t *bamHdr = sam_hdr_read(fp_in); //read header

	printf("%s\n", bamHdr->target_name[1]);

	return 0;
}

$ ./s3_htslib s3://test-bucket/test.bam
2
$
```

Would you mind double-checking that the new `build.rs` does not exclude `hfile_s3.c` symbols from `libhts.a` or similar? Perhaps something is off on the Rust/bindgen side instead?

brainstorm:
Just noticed that the `static = []` change/feature seems to be broken? From my crate's code after selecting `features=["s3", "static"]`:

```
$ cross build --release --target x86_64-unknown-linux-musl
error: failed to select a version for `rust-htslib`.
    ... required by package `reads v0.1.0 (/project/reads)`
versions that meet the requirements `= 0.30.1-alpha.0` are: 0.30.1-alpha.0

the package `reads` depends on `rust-htslib`, with features: `static` but `rust-htslib` does not have these features.


failed to select a version for `rust-htslib` which could resolve this conflict```

**Comment by pmarks** [comment]
@brainstorm - htslib is causing curl 7.70 to generate err = CURLE_UNKNOWN_OPTION while setting an FTP option here:
err |= curl_easy_setopt(fp->easy, CURLOPT_FTP_FILEMETHOD, CURLFTPMETHOD_NOCWD);

https://github.com/rust-bio/htslib/blob/4a913bad6866ba17800ec22b143246aab007ef9c/hfile_libcurl.c#L1173

This is because the `curl-sys` crate doesn't build curl with FTP support by default. I add the `protocol-ftp` feature to curl-sys, then it  makes a https request to s3.  I guess it's OK to add FTP support to curl?

I'm not getting a 403 back from sw, which is the same as what happens in my browser.  Do you have crdentials for that test bucket that we can put into a test?

**Comment by pmarks** [comment]
@brainstorm I added a simple s3 test to the bucket you mentioned - can you add creds?
I also tried with this file: http://s3.amazonaws.com/1000genomes/1000G_2504_high_coverage/data/ERR3239276/NA06985.final.cram

I can get it with firefox, but I'm getting access when I put `"s3://1000genomes/1000G_2504_high_coverage/data/ERR3239276/NA06985.final.cram"` into the test.

**Comment by brainstorm** [comment]
TIL: Nice, I would have never imagined that S3 would actually require FTP bits, always thought of it as HTTP-only protocol, nice forensics @pmarks 👍 

I was testing with an internal private bucket/object so I cannot publish that, so let's stick with the public 1000genomes one?

Unfortunately there's still [something else off](https://github.com/brainstorm/rust-htslib/commit/33f258988328670dc1236c28379fc65cc7d35cd0#diff-48d1ec63cbd733608f19d3a6be63dd4bR1819) after activating FTP support on libcurl:

```rust
$ cross test --release --features="s3" --target x86_64-unknown-linux-musl
(... all tests ok until...)
[E::hts_open_format] Failed to open file "s3://1000genomes/1000G_2504_high_coverage/additional_698_related/data/ERR3989458/NA20358.final.cram" : Input/output error
test bam::tests::test_s3_connect ... FAILED

failures:

---- bam::tests::test_s3_connect stdout ----
Err(
    Open {
        target: "s3://1000genomes/1000G_2504_high_coverage/additional_698_related/data/ERR3989458/NA20358.final.cram",
    },
)
thread 'bam::tests::test_s3_connect' panicked at 'called `Result::unwrap()` on an `Err` value: Open { target: "s3://1000genomes/1000G_2504_high_coverage/additional_698_related/data/ERR3989458/NA20358.final.cram" }', src/bam/mod.rs:1823:18


failures:
    bam::tests::test_s3_connect

test result: FAILED. 74 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '--lib'
```

Same with MUSL:

```
$ cross test --release --features="s3" --target x86_64-unknown-linux-musl
[E::hts_open_format] Failed to open file "s3://1000genomes/1000G_2504_high_coverage/additional_698_related/data/ERR3989458/NA20358.final.cram" : I/O error
```

brainstorm:
@pmarks Found a suitable S3-BAM public dataset:

s3://gatk-test-data/wgs_bam/NA12878_24RG_hg38/NA12878_24RG_small.hg38.bam

Same effect with BAM though: `Input/output error`, debugging...

pmarks:
@brainstorm ok I'm making prebuilt bindings the default. I added a note to the README about it.

I added a test to access a BAM file via HTTP.  I think the s3 support is basically there, but I think there's various details about authentication that the user needs to sort out with env vars and url params. `samtools view` seems to have the same issue.  I put a note in the README aobut s3 & gcs being 'beta-level'.  I'd propose we merge this and keep working on testing s3/gcs later.

brainstorm:
100% agree, merge this, S3 support can come right after, thanks @pmarks, great work here indeed!

brainstorm:
> samtools view seems to have the same issue

Weird, it works fine for me (at least just printing the BAM header:

```
$ samtools view -H s3://gatk-test-data/wgs_bam/NA12878_24RG_hg38/NA12878_24RG_small.hg38.bam

(...)
@SQ	SN:chrUn_JTFH01001935v1_decoy	LN:2330	AS:38	AH:*	M5:bb4702fc9e6227387e8ce3498109d998	UR:/seq/references/Homo_sapiens_assembly38/v0/Homo_sapiens_assembly38.fasta	SP:Homo sapiens
@SQ	SN:chrUn_JTFH01001936v1_decoy	LN:2327	AS:38	AH:*	M5:ddad50424c1114dd98bc0202ac1cd9fa	UR:/seq/references/Homo_sapiens_assembly38/v0/Homo_sapiens_assembly38.fasta	SP:Homo sapiens
@SQ	SN:chrUn_JTFH01001937v1_decoy	LN:2318	AS:38	AH:*	M5:28804e3d50bf7764ac6b1897b1128790	UR:/seq/references/Homo_sapiens_assembly38/v0/Homo_sapiens_assembly38.fasta	SP:Homo sapiens
(...)
```

It does work as well for private buckets that need priv/pubkeys....anyway, we'll get to the bottom of this in any case ;)

landesfeind:
Hi @pmarks, are the `faidx.c` missing in the prebuilt bindings? How can we/I add them?

Context: Pull request #214 adds faidx bindings and compilation works well with `cargo build --features bindgen` but fails without the bindgen feature. I diff'ed the prebuilt bindings with the ones from a fresh build but there are many differences. Thus I guess, it's not a simple `cp target/release/.../bindings.rs linux_prebuilt_bindings.rs`.

PS: I like #189 since I had quite some trouble with building htslib on a CentOS with old bindgen :-)

brainstorm:
@pmarks I believe I'm facing a similar issue with the bindgen prebuilt bindings that Manuel found here... could you [document the prebuilding of bindgen](https://github.com/rust-bio/rust-htslib/pull/214#issuecomment-644381026) a bit more thoroughly? I tried to just `cargo build --features bindgen` on my OSX machine as pointed above but the resulting `bindings.rs` on the `target/release` folder does not look correct (i.e: a lot of threadpool bindings are deleted which I guess they shouldn't?):

```
(...)
-extern "C" {
-    pub fn hts_tpool_process_attach(p: *mut hts_tpool, q: *mut hts_tpool_process);
-}
-extern "C" {
-    pub fn hts_tpool_process_detach(p: *mut hts_tpool, q: *mut hts_tpool_process);
-}
-extern "C" {
-    pub fn hts_tpool_process_ref_incr(q: *mut hts_tpool_process);
-}
-extern "C" {
-    pub fn hts_tpool_process_ref_decr(q: *mut hts_tpool_process);
-}
```

Pregenerating a correct `osx_prebuilt_bindings.rs` bindgen should hopefully be the last thing needed to address #275.

PR #191, opened 2020-03-10
enable musl builds

johanneskoester:

PR #193, opened 2020-03-19
Musl support for htslib 1.10.2

brainstorm:
This is a followup on https://github.com/rust-bio/rust-htslib/pull/184 and https://github.com/rust-bio/rust-htslib/pull/191 to restoring musl support for htslib 1.10.2 now that it requires openssl and curl (because of hfile_s3 et al).


brainstorm:
@nlhepler @johanneskoester Looks like I'm at the last linking step and `hts-sys` goes into gcc's `ld` when it should go for `/usr/local/musl/bin/x86_64-linux-musl-ld` instead:

https://github.com/rust-bio/rust-htslib/pull/193/checks?check_run_id=518274763#step:5:252

Do you have any good advice to fix that last thing up? [The container](https://github.com/rust-bio/rust-htslib/pull/193/files#diff-8ad8e65f38db4892ff5e8a884e95ec25R2) has a full musl-cross-compiled toolchain with all the deps needed and I could compile `hts-sys/htslib` manually from within the container... I thought about just going like that towards the end of the building phase for the container:

```
# rm /usr/bin/ld && ln -sf /usr/local/musl/bin/x86_64-linux-musl-ld /usr/bin/ld
```

But I'm pretty sure there are much better ways to get this fixed properly? Any `build.rs` variable/flag/option that would force using the `musl`'s `ld` binary instead of the system's one?

brainstorm:
Now I [managed to compile all the way up to the `wrapper.h`](https://github.com/rust-bio/rust-htslib/runs/523847455?check_suite_focus=true#step:5:332), looking into ways to pass `-I/usr/local/musl/x86_64-linux-musl/include` flags to rust bindgen...

/cc @pmarks @nlhepler

brainstorm:
@johanneskoester [Seems to work now](https://github.com/rust-bio/rust-htslib/runs/524620170), but linting is happening outside `cross` containers... would it be ok for you to move the linting inside `cross` instead of the host system?

brainstorm:
Now it works for "all features enabled", but it dies for the `--no-default-features` variant:

https://github.com/rust-bio/rust-htslib/runs/544923906?check_suite_focus=true#step:9:261

johanneskoester:
Thanks a lot! I have little knowledge about musl and so on, so just let me know when you feel like being ready for merging!

hts-sys/build.rs
```diff
@@ -93,6 +106,7 @@ fn main() {
         .generate_comments(false)
         .blacklist_function("strtold")
         .blacklist_type("max_align_t")
+        .clang_arg("-I/usr/local/musl/x86_64-linux-musl/include")
```
pmarks, line 109: Maybe make this line conditional on being in the musl target?

brainstorm: Here here, I wanted to throw away the current makeshift container in favour of Alpine linux (which is MUSL-friendly by default) so that arg would not be needed at all, but then that alpine container went "sh: 1: cargo: not found" on me, need to dig a bit further to fix the alpine container now :/

brainstorm:
Thanks @johanneskoester @pmarks! I just need a little bit of help with  [`--no-default-features`](https://github.com/rust-bio/rust-htslib/pull/193/files#diff-cf310bd8aed83600b1472d2ea1ec4290L84) (currently disabled on GitHub Actions) and this PR should be mergeable... To be honest I'm not entirely sure how the "--no-default-features" should function since I guess that `zlib` is a fairly basic "feature", but the current error and hts-sys implementation suggests that the lib can be removed? I'm a bit confused.

pmarks:
@brainstorm - here's a guess: `curl-sys` depends on `libz-sys` which is what's providing the `DEP_Z_INCLUDE` env var. So if we add `libz-sys` to the deps of `hts-sys` then it should work.  More discussion here:
https://doc.rust-lang.org/cargo/reference/build-script-examples.html

hts-sys/Cargo.toml
```diff
@@ -17,15 +17,16 @@ pre-release-commit-message = "release version {{version}}"
 tag-message = "Version {{version}} of Rust-HTSlib."
 
 [dependencies]
-libc = "0.2"
-libz-sys = "1.0"
+libc = "0.2.68"
+libz-sys = { version = "1.0.25", optional = true }
```
pmarks, line 21: I don't think libz-sys should be optional.  It's required for BAM and vcf.gz support and there's no way to turn off those features in hts-lib, so the user shouldn't be allowed to disable that dependency.

brainstorm:
Sorry for the failed CI noise (if you receive the email notifications), I'm back with this work and closing in to a solution with a static-friendly version of `build.rs` after posting this yesterday:

https://users.rust-lang.org/t/static-builds-fpic-cc-build-flag-ineffective/42586/2

In a nutshell, I suspect that that `hts-sys`'s build.rs is passing DSO linker flags to dependant packages (htsget-aws::reads in my case), even when the desired outcome should be `static` (in my case/interests)... WIP over here for a rough idea on what I mean:

https://github.com/brainstorm/rust-htslib/commit/d8d1c5a8dddf01bb5aba24468bbc444f7035155b#diff-354901c3ea564c6ea68f8af7eccdacdbR34

/cc @jkbonfield @daviesrob @ohofmann

brainstorm:
The last couple of commits seem to have addressed most of the problems with the static flags for the different dependencies and `-fPIC` being passed on `build.rs` to htslib's build system.

Now the issue is only with functions that I use directly on my crate (`sam_hdr_read`, `sam_index_load`, etc), see:

https://users.rust-lang.org/t/static-builds-fpic-cc-build-flag-ineffective/42586/2

I suspect that this has something to do with rust-htslib's wrapper now perhaps? I'll keep investigating but I would very much welcome any hints @johanneskoester @pmarks at this point :)

brainstorm:
Nevermind @johanneskoester @pmarks, fixed, it's working now but hts-sys needs a good cleanup to support both static and dynamic building.

brainstorm:
@pmarks To your comment on https://github.com/rust-bio/rust-htslib/pull/193#issuecomment-609846824 ... I might be wrong or overcomplicating this, but I suspect it's a bit more involved than that, see:

https://users.rust-lang.org/t/musl-cross-compilation-how-to-reuse-curl-sys-libz-sys-bzip2-sys-lzma-sys-and-other-dependency-includes-under-target/42817/3

brainstorm:
@johanneskoester Did you change any of the GitHub actions configuration for this repo recently? I'm now getting the following when pushing to this branch (yesterday it was fine):

```
(base) [limsadmin@5180-novastor01 rust-htslib]$ git push
Counting objects: 28, done.
Delta compression using up to 40 threads.
Compressing objects: 100% (14/14), done.
Writing objects: 100% (20/20), 2.11 KiB | 0 bytes/s, done.
Total 20 (delta 10), reused 0 (delta 0)
remote: Resolving deltas: 100% (10/10), completed with 6 local objects.
To https://github.com/brainstorm/rust-htslib.git
 ! [remote rejected] musl_support -> musl_support (refusing to allow an OAuth App to create or update workflow `.github/workflows/rust.yml` without `workflow` scope)
```

brainstorm:
@johanneskoester Both MUSL and GNU are now working thanks to Cross and the docker containers under `docker/`... unfortunately I cannot push any changes to `.github/workflows/rust.yml` anymore, so this PR will be stalled until this gets addressed, unfortunately :-S

brainstorm:
It turns out that the errors above where caused by having a HTTPS endpoint (which requires OAuth tokens now for Github Actions) instead of SSH. Switched to `git://` and it's rolling again.

hts-sys/build.rs
brainstorm: @pmarks Fixed with the new containers, although I might need to reintroduce this `.clang_arg` since CI builds do not pass without https://github.com/rust-bio/rust-htslib/pull/193/commits/b32524b84cbd6620db5fcff4d99ff64176cdd838 in place.

Thoughts?

johanneskoester:
Awesome, ready for merging?

brainstorm:
@johanneskoester It is! The last small bit to "solve" would be to put `CFLAGS=-I/usr/local/include/musl` as part of the `build.rs` script depending on `cc`'s target (only when MUSL, as @pmarks pointed out), but it wasn't clear to me from the `cc` docs how to query that. 

Can be solved post-merge though since it's clearly mentioned on the README.md, I hope it doesn't generate too many issues from users.

Merging then ;)

PR #203, opened 2020-05-22
restrict rerun-if-changed to build files

nlhepler:
especially to ignore the .git directory, which we've noticed may be contributing to constitutive rebuilds of hts-sys in some circumstances.


nlhepler:
Not sure why coverage upload failed, maybe the github outage from last week? re-running and possibly rebasing afterwards if it continues to fail.

brainstorm:
Thanks @nlhepler I did notice some issues with `hts-sys` crate too across rebuilds.

PR #208, opened 2020-06-05
libdeflate, cloudflare's zlib, versioned containers

brainstorm:
* Introduce libdeflate to both containers.
* Introduce [cloudflare's zlib](https://blog.cloudflare.com/cloudflare-fights-cancer/) for a claimed 20% performance boost.
* Be a responsible netizen, use semver for containers instead of `latest` ;)
* Inspired by Brent Pedersen's hts-nim: https://github.com/brentp/hts-nim/blob/master/Dockerfile.musl-hts-nim


juliangehring:
Nice! Just for reference, some benchmarks that show the benefits for htslib: http://www.htslib.org/benchmarks/zlib.html

brainstorm:
Thanks @juliangehring for the benchmarks pointer, interesting!

@pmarks Would you mind factoring `libdeflate` in while working on https://github.com/rust-bio/rust-htslib/pull/207#issuecomment-639211730? I'm not entirely sure if it pays off to have it optional or we could just ship it enabled by default (see benchmarks)?

I very much trust your judgement more than mine since I'm new around here, so happy to adapt ;)

Cheers!

pmarks:
@brainstorm cool! See my PR - I'm trying to refactor how we build `htslib` to give us more control of the options without relying on `./configure`.  `libdeflater` should fold into that nicely, I think that crate just needs some directives added to their `build.rs`.

brainstorm:
Gotcha, I'll leave this PR only with the docker changes then, good work Patrick!

Issue #209, opened 2020-06-06
64 bit BAM-tag values are written to 32 bit fields


jch-13:
According to the BAM specs (https://samtools.github.io/hts-specs/), numeric values in auxiliary fields ("tags") are stored as 32 bit values. However, `htslib` returns 64 bit values when _reading_ those fields, even though they are stored as 32 bit values (`int64_t bam_aux2i(...)` and `double bam_aux2f(...)`), which is probably the reason why `rust-htslib`'s  `Aux::Integer(i64)` and `Aux::Float(f64)` enum representations of those fields correspondingly wrap 64 bit values.
https://github.com/rust-bio/rust-htslib/blob/3fa020da8b04de5d8323f1bc0e6332f3bf154408/src/bam/record.rs#L762-L769

Hence, when _writing_ BAM tags with the `bam::Record::push_aux(..., value: &Aux<'_>)` method, 64 bit values get written to 32 bit fields (quite explicitly by setting the offset to 4 bytes in the `push_aux(...)` method), which leads to misbehavior.
https://github.com/rust-bio/rust-htslib/blob/3fa020da8b04de5d8323f1bc0e6332f3bf154408/src/bam/record.rs#L601-L619

As far as I can see, `htslib` supports a `d` (double) type to store 64 bit floating point values (which I cannot find in the specs), but there does not seem to be a 64 bit equivalent for integers.

Edit: The float case can simply be solved by adding a `Aux::Double(f64)` variant and changing the existing `Aux::Float(f64)` to `Aux::Float(f32)` (API-breaking), but I'm not quite sure how to approach the integer case ergonomically (and preferably checked at compile-time). I think none of these approaches are overly satisfying:
- Adding a 32 bit enum variant and reject 64 bit values at runtime in `push_aux(...)`
- Changing the enum variant to `Aux::Integer(i32)` and down-cast automatically when reading BAM tags
- Adding another enum just for writing BAM tags

dlaehnemann:
Good catch. And this does look tricky to fix. From what I can see, the following functions are the ones publicly available for first getting the pointer to an AUX tag's data and then getting out an integer value (all the other possible types follow below, but I skip them to keep the code rendered here to a minimum):

https://github.com/samtools/htslib/blob/develop/htslib/sam.h#L1357-1376

So, `bam_aux2i` will always return an `int64_t`, no matter what the original type of the stored data was (`i8`, `i16`, ...). How this happens, becomes clear from this htslib-internal code:

[...]

PR #210, opened 2020-06-10
Use `std` instead of `itertools` for collecting vectors

juliangehring:
`itertools` is only used to collect into vectors. We can use `std` instead.


brainstorm:
Thanks @juliangehring and @jch-13 for spotting this, nice forensics on https://github.com/rust-bio/rust-htslib/issues/209.

PR #212, opened 2020-06-13
Make tests more idiomatic

juliangehring:
This makes some tests a bit simpler and more idiomatic. It also addresses many clippy warnings. Only code in tests is modified, there are no breaking changes to the user of the library.

I have left two clippy warnings about complex patterns unaddressed since I don't think "fixing" them would be an improvement.


src/bam/ext.rs
```diff
@@ -9132,7 +9132,7 @@ mod tests {
         }
 
         let mut read = bam::Record::new();
-        for (input_cigar, supposed_length_wo_hard_clip, supposed_length_with_hard_clip) in vec![
```
brainstorm, line 9135: @juliangehring I had no idea `&` (slice?) was considered more idiomatic than the `vec!` macro (which I consider easier to read as a dev), but I trust you ;) 

I'll merge this all once it goes green after the master branch merge.

juliangehring: I think the idea is to prefer a slice when a vector is not needed. In any way, at least it makes clippy happier ;)

PR #214, opened 2020-06-14
Add basic functionality to read indexed FASTA files (continued)

landesfeind:
This pull request supersedes the #165 in using an own branch for easier integration. Most changes of @dlaehnemann (see #165 review) are incorporated.


dlaehnemann:
To make the clippy errors go away and make this pull request work, I think you have to add `#include "htslib/htslib/faidx.h"` to the include list at the start of `hts-sys/wrapper.h`:

https://github.com/rust-bio/rust-htslib/blob/61b0644b73ab40612cd23e104ff0a3fe647bf064/hts-sys/wrapper.h#L1-L9

landesfeind:
Hm, that one is actually there: https://github.com/landesfeind/rust-htslib/blob/faidx/hts-sys/wrapper.h#L10
Also, I ran the full CI pipeline in my own repository before I submitted the pull-request: https://github.com/landesfeind/rust-htslib/actions/runs/135013019

Let's try with a merging of the master branch.

dlaehnemann:
You're right. Overlooked that, as well... But it still can't find the relevant structs and functions, weird. I guess merging in the master is always worth a try...

landesfeind:
Do the `*_prebuilt_bindings.rs` probably miss faidx? 
After merging the master, also my local build fails using `cargo build`. But it works well when enforcing generation of the bindings using `cargo build --features bindgen`.
Any chance you can give me a hint how the prebuilt are update? I compared the `bindings.rs` generated during my "bindgen-build" with the prebuilt one but there are many many differences. I think I can not simply copy them over.

dlaehnemann:
From looking at it, this does seem like the problem. Check out the new info at the end of the [README.md section on Usage of rust-htslib](https://github.com/rust-bio/rust-htslib#usage). It's because using `bindgen` is now an optional feature, which requires prebuilt bindings for all functionality that is to work without bindgen. For context, you could checkout PR #189 which was merged 3 days ago.
But it looks like for the faidx functions and structs, this can easily be modelled upon existing code, right? Otherwise just ping in pmarks and brainstorm, they should know much more about this than I do.

pmarks:
@landesfeind I will add updated bindings to include faidx stuff & send you a branch that you can merge into this.  (I'll also include instructions for how to do the update).

pmarks:
@landesfeind try merging my branch [here](https://github.com/rust-bio/rust-htslib/tree/pmarks/faidx-bindings) into this branch -- that should fix build without bindgen.

landesfeind:
Hi @pmarks, that did the trick! Thanks a lot! Ready for merge, unless there are more ideas on code improvement :-)

.github/workflows/rust.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, faidx ]
```
brainstorm, line 5: This shouldn't be needed when it gets merged to master ;)

landesfeind: Thanks for noting!  :)

src/faidx/errors.rs
```diff
@@ -0,0 +1,13 @@
+use snafu::Snafu;
```
johanneskoester, line 1: I am super sorry, but I just migrated error handling from snafu to thiserror (as it seems to become quite standard nowadays). Can you adapt it here?

src/faidx/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /// Fetches the sequence and returns it.
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - the name of the template sequence (e.g., "chr1")
+    /// * `begin` - the offset within the template sequence (starting with 0)
+    /// * `end` - the end position to return (if smaller than `begin`, the behavior is undefined).
+    pub fn fetch_seq<N: AsRef<str>>(&self, name: N, begin: usize, end: usize) -> String {
```
johanneskoester, line 77: I think the bounds should be u64. usize can be 32 bit on 32 bit machines, which might be not enough to index into a large fasta file.

johanneskoester: Is there a particular reason to return a unicode string? Why not just bytes?

johanneskoester: Also, I think it might make sense to let people provide a buffer, instead of returning a newly allocated string every time.

landesfeind: Hi Johannes, thanks a lot for the suggestions!

> I think the bounds should be u64. usize can be 32 bit on 32 bit machines, which might be not enough to index into a large fasta file.

Valid point. The internal htslib type is c_int (=i32). Either we use that directly or switch to `faidx_fetch_seq64()` using hts_pos_t (=i64). I would suggest the latter but its your call :-)

> Is there a particular reason to return a unicode string? Why not just bytes?

Because I like Strings over raw byte arrays (because I am a simple mind and did not need optimize my code for speed so far). But clearly can change that. I will implement this along with a pre-allocated buffer. Maybe I that as separate functions (i.e., syntactic sugar). Just give me a couple days, because I have a busy week.

landesfeind: Hi @johanneskoester, 1770a09e437558a96cce4bbcc2eab2c011d8b96e updates the branch to use i64 type as parameter. This is consistent with the htslib C-API that uses [hts_pos_t](https://github.com/samtools/htslib/blob/835e133b941a175246b18a1a3261f7d91611a1dc/faidx.c#L695) which is [defined as i64](https://github.com/samtools/htslib/blob/34c15969fa49d78b07c649374fb008fb3ed1e98a/htslib/hts.h#L629). I agree that `u64` would be nicer but we then would need to check overflows and either return a `Result` or `panic` - both options I do not like.

I changed the return value to byte array. Providing a buffer is not useful because the C-API creates a new byte array anyways and [converting CStr to bytes should be computationally cheap](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.to_bytes).

brainstorm: Why is `Result` not acceptable as a return type?

landesfeind: `Result` is a perfectly acceptable return type in general but in this case the single thing that can fail is the cast. A developer using this API needs take care either before (i.e., doing the cast by him/herself) or after (i.e., matching the Result) the call. Therefore, I consider returning a `Result` here as a kind of overhead.

I also think that @lh3 (and/or htslib-team) had a reason to choose `i64` as type for positions. Imho the `rust-htslib` wrappers should resemble the `C-htslib` API closely. Even if I don't know what `C-htslib` will do if a negative value is supplied, one day a developer might want to use the `rust-htslib` in that way.

brainstorm: See https://github.com/samtools/hts-specs/pull/460 for a good discussion on signedness (for .BAI format in this case but the rationale should apply to the rest of positional-encoding fields).

TL;DR: positions should be positive on the current spec, if they are not, it's probably a bug in a spec that hasn't seen proper scrutiny in ~10 years ;)

landesfeind: Hi @brainstorm, 
I agree that making the API ergonomic for Rust programmers is a good goal. Also, I was not aware of the discussion in hts-specs and naively assumed that i64 was used for a specific reason. I now switched to `usize` type and added an Error type for in case conversion fails. Not sure though if the description gives enough detail:
```
#[derive(Error, Debug, PartialEq)]
pub enum Error {
   ...
    #[error("The given position is to large to be converted to i64")]
    PositionTooLarge,
}

File: src/faidx/errors.rs [hunk 4]
```diff
@@ -0,0 +1,14 @@
...
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[derive(Error, Debug, PartialEq)]
+pub enum Error {
+    #[error("file not found: {path}")]
+    FileNotFound { path: PathBuf },
+    #[error("invalid (non-unicode) characters in path")]
+    NonUnicodePath,
+    #[error("The given position is to large to be converted to i64")]
```
brainstorm, line 12: Typo: "to large->too large"

landesfeind: Oh - thanks a lot for noticing!

brainstorm:
Thanks for your contribution @landesfeind

landesfeind:
You are welcome - thank you all for the feedback!

PR #219, opened 2020-06-22
Fix `libdeflate` feature selection

juliangehring:
The `libdeflate` feature (added in #189) isn't currently propagated and the feature gets silently ignored during build. Fixing the feature names and dependencies addresses this, and also ensures that the feature-specific tests are correctly ignored.

If someone knows a good way to automatically test that the correct deflate library is included, I'd be happy to add this as well.


brainstorm:
Good catch @juliangehring, thanks!

> If someone knows a good way to automatically test that the correct deflate library is included, I'd be happy to add this as well.

That should be tackled by libdeflate-sys itself afaik (it'll error out if not found or sth else is wrong)? Regardless, let's merge this PR, looks fine and seems to work now (local test), thanks again!

brainstorm:
Oh, sorry, you meant testing as in `use libflate` and assert some result on tests?

Issue #194, opened 2020-03-31
tbx::Reader can no longer read indexed text files with version 0.27+: "Error: invalid tabix index"


felix-clark:
I am trying to use the `tbx::Reader` read an indexed ~~BED~~ text file with data lines of the form
```
<chrom>    <pos>    <data...>
```
that have `.tbi` index files generated with `tabix -b2 -e2 <file>.bed.gz`. Calling `tbx::Reader::from_path()` works as expected in verion 0.26, but generates an error "invalid tabix index" in 0.27 and later. ~~Version 0.27 of `rust-htslib` changes the type of the index from 32 bits to 64 bits, presumably to keep up with the changes in HTSlib. (this is unrelated)~~ I have tried using a recent version of tabix (1.10.2) and re-indexing the file, but this did not solve the issue (and in fact the resulting `.tbi` index file appears to be identical).

I apologize for not having the time to dive into this further myself right now, and it's possible that the underlying HTSlib C library is ultimately at fault, but I thought I'd create the issue now in case someone else is experiencing it.

juliangehring:
@felix-clark I tried reproducing this with a simple data set

```
chr7    127471196       1
chr7    127472363       2
chr7    127473530       3
chr7    127474697       4
chr7    127475864       5
```

and tabix 1.10.2, but didn't see any errors with v0.26.0 to v0.30.0.

Is there a data set which gives you the error and that you can share for reproducing this issue?

felix-clark:
@juliangehring It seems like commented header lines are responsible. Perhaps you could  prepend that data with a line like `#contig pos     data`?

juliangehring:
Okay, the header line makes all the difference - I can reproduce it when the file has a header line.

It seems that this is caused by the file format recognition in htslib. Without the header, it recognises the file format category as a region list. With the header, it classifies it as a general text file and puts in the unknown file format category - this is why the the safety checks in `Reader::new` fail and the `InvalidIndex` is raised.

Maybe the safety checks for the correct file formats can be extended a bit to support more formats, including files that are not properly recognised. However, it ultimately looks like a bug in htslib to me.

felix-clark:
This doesn't appear to have been flagged as an issue in htslib in the time since the last discussion here, but it does seem like something that should ideally be addressed upstream.

I submitted a proposed PR here to permit this particular case along with some tests and test files, both positive and negative. It should definitely be reviewed by an expert.

felix-clark:
[...]

PR #226, opened 2020-07-23
Tabix reader fix for custom text files

felix-clark:
An update to htslib broke the ability to recognize the format of simple ~~BED~~ text files with a header line (#194). This PR adds a special case to the file format check to permit the `tbx::Reader` to attempt to read these files.

Some tests are added to ensure this format is recognized, including one with a nonsense index file that is expected to return an error.


felix-clark:
The failed test is due to a MUSL build, but it doesn't look directly related to this PR. Let me know if that assumption is incorrect.

src/tbx/mod.rs
```diff
@@ -151,16 +151,22 @@ impl Reader {
...
-            if (*hts_file).format.category != htslib::htsFormatCategory_region_list
-                && (*hts_file).format.format != htslib::htsExactFormat_sam
-            {
-                return Err(Error::InvalidIndex);
-            }
+        let file_format = unsafe { (*hts_file).format.format };
+        let format_category = unsafe { (*hts_file).format.category };
+        match (format_category, file_format) {
+            (htslib::htsFormatCategory_region_list, _) => (),
+            (_, htslib::htsExactFormat_sam) => (),
```
felix-clark, line 158: I wonder if this line permitting "sam" format is no longer needed, based on the htslib changelog here:
https://github.com/samtools/htslib/blob/82d01ba63b20eed8905ea60c2e1cfe0f96c73bab/NEWS#L222
The tests pass if it is removed; is that sufficient to conclude it's no longer needed?

src/tbx/mod.rs
```diff
@@ -151,16 +151,22 @@ impl Reader {
...
+            // simple BED file with a header line. See #194
+            (htslib::htsFormatCategory_unknown_category, htslib::htsExactFormat_text_format) => (),
+            _ => return Err(Error::InvalidIndex),
         }
 
         let hts_format = unsafe { (*htslib::hts_get_format(hts_file)).format };
         let tbx = unsafe { htslib::tbx_index_load(path.as_ptr()) };
+        if tbx.is_null() {
+            return Err(Error::InvalidIndex);
+        }
```
felix-clark, line 169: The check for nullity is moved earlier to prevent a possible null dereference in the block below.

src/tbx/mod.rs
felix-clark: Discussion on #194 confirms that this interpretation is correct, and also suggests that this entire check may be unnecessary.

felix-clark:
It's been suggested that the file format check can be removed entirely. Waiting on additional discussion in the issue before proceeded.

brainstorm:
What was the actual suggestion? Looks like all messages from John Marshall were removed? :-S

felix-clark:
@brainstorm The suggestion was to consider removing the check on the file format and format category entirely. I'm not an expert but this does smell reasonable to me, since the tabix reader should probably try to read any file that has been successfully indexed in the first place. Unfortunately I have no clue why the check existed in the first place.

For what it's worth, the tests all pass when the check is removed, including one I added in this PR that verifies that trying to read a data file with a nonsense index file returns an error.

brainstorm:
I would remove those redundant checks then? From what you tell me and see in the code there's little need for "double-gating" format detection on both htslib **and** rust-htslib, right?... carry on ;)

felix-clark:
Thanks for the attention, @brainstorm . I've removed this check in the latest commit.

src/tbx/mod.rs
```diff
@@ -422,4 +417,16 @@ mod tests {
...
+    fn test_bed_headered() {
+        // This file has chromosome, start, and end positions with a header line.
+        Reader::from_path("test/test_bed4.bed.gz").expect("Error opening file.");
+        // This file has chromosome and position with a header line, indexed with
+        // `tabix -b2 -e2 <file>`.
+        Reader::from_path("test/test_bed5.bed.gz").expect("Error opening file.");
+        // This is a duplicate of the above file but the index file is nonsense text.
+        Reader::from_path("test/test_bed_bad_index.bed.gz")
+            .expect_err("Invalid index file should fail.");
+    }
```
dlaehnemann, line 431: I really like those test cases. But to avoid any future confusion, I think we should only call any file a bed file that conforms to (what pretty much amounts to) the specification of BED files here:
https://www.genome.ucsc.edu/FAQ/FAQformat.html#format1

This only applies to `test/test_bed3.bed.gz` (which I would add as another test case, here;). For `test/test_bed[45].bed.gz`, we have an unallowed type of header line as the UCSC definition linked above only allows `track` header lines. And for `test/test_bed5.bed.gz` we miss the third required column (`end`) for this file to be called a bed file. For tabix indexing, this is fine, but I would simply call them something else (e.g. `4` a `genomic_regions.txt` file and `5` a `genomic_positions.txt` file?).

And another very little change I would suggest, is to move all the test files for this reader into a subfolder in test, e.g. `test/test_tabix_reader/`. This way we don't create one huge folder of test files for all kinds of different functionality as we add tests...

src/tbx/mod.rs
dlaehnemann: Jep, moving this up here makes this safer and removing the above file type checks should be perfectly fine. Using `htslib`-functionality like `hts_get_format` should always be the preferred way of doing this and in the future, I guess we should try to document any cases where we deviate from this with a quick code comment stating why...

dlaehnemann (approved):
Thanks for keeping at this! It looks good to me. I would clean up the tests a bit further (as suggested) and then this should be good to merge.

In the future, we might want some doctests in here, as well, especially for all the functions that are not really tested in detail, right now. But I guess that this is something to address from a soon-to-be planned Docathon onwards... ;)

dlaehnemann:
Also, I think it should be fine to simply add the suggested changes to your branch and we can then update the branch from `master` right here in the PR (i.e. online).

src/tbx/mod.rs
felix-clark: Great suggestions! I've followed them in cee3c88, with the exception of including `test_bed3.bed.gz` in these tests, since it is covered at a basic level elsewhere in this test module. I also split this function into 3 separate test functions (one for each file) so if they fail in the future it is easy to tell which one is problematic.

felix-clark:
Thanks for the review, @dlaehnemann. The test CI check is failing only due to `bam::tests::test_http_connect`, for which I also noticed intermittent failures when testing locally. Assuming that can be ignored, we should be good to go.

src/tbx/mod.rs
dlaehnemann: Ah, OK. I simply didn't see that bed3 test in the changes...

dlaehnemann:
I reran the CI and it passed this time around. So I'll merge this, thanks again!

PR #232, opened 2020-08-07
Calculate test coverage with all features

juliangehring:
Includes all features in the calculation of the test coverage. Before, only the default features were used which misses some relevant code parts (e.g. the `serde` feature).


juliangehring:
Using all features raises the total test coverage to [93.8%](https://coveralls.io/builds/32604822).

PR #231, opened 2020-08-06
Musl compile fix

brainstorm:
Seems like recent changes on the underlying musl/gcc toolchains affecting DWARF trigger build failures for debug (non `--release` builds).

For more context, the current [rust-musl-make](https://github.com/richfelker/musl-cross-make) binutils version is 2.33.1, which apparently has compression for the debug symbols. see, i.e: https://github.com/mrjbq7/ta-lib/issues/268#issuecomment-497979600

I don't like to downgrade toolchain versions either, so I'll make the assumption that debug builds for MUSL are not very used ATM... my workflow is (re-)building and deploying AWS lambdas in `--release` mode for now.

PR #233, opened 2020-08-12
Slim docker musl

brainstorm:
Solves the debugging issues (non `--release` builds breaking because of compressed debug symbols).... as observed in https://github.com/rust-bio/rust-htslib/pull/231. 

Also general cleanup, reducing final docker image size. Further cleanups expected if https://github.com/rust-embedded/cross/pull/452 passes and gets merged in, otherwise we'll be using `brainstorm/cross-x86_64-unknown-linux-musl:upstream` for now.

~The cloudflare speedups on zlib will have to wait until https://github.com/rust-lang/libz-sys/issues/63#issuecomment-672687388 gets approved/repointed.~ ... I'll experiment with this crate in a separate PR:

https://crates.io/crates/cloudflare-zlib-sys


juliangehring:
@brainstorm Can the `libdeflate` feature still be used when it gets removed in the [`Dockerfile.gnu`](https://github.com/rust-bio/rust-htslib/pull/233/files#diff-a1ce3ee37bfffa3807fe76a732ea3899L9-L11) or is it blocked by the issue as the cloudflare-based compression?

brainstorm:
@juliangehring Two different beasts: [`libdeflate`](https://crates.io/crates/libdeflater) is a separate crate than [`zlib`](https://crates.io/crates/libz-sys)... but thanks to your question I found this crate: https://crates.io/crates/cloudflare-zlib-sys ... for which I'll be pullrequesting soon ;)

To your question: yes, `libdeflate` should be pulled by its `-sys` crate automatically, so no previous system install should be needed, AFAICT.

Other than that, I think this PR is good to merge.

PR #234, opened 2020-08-12
Let buffer return Arc.

johanneskoester:


brainstorm:
Isn't `Arc` more expensive than `Rc`? Why is this needed? :)

johanneskoester:
This way, it is possible to easily parallelize computations on the returned Records. Other options would be to make this somehow optional, or wrap the Rc in an Arc for parallel computations. Do you have a feeling how much overhead it adds?

veldsla:
Quick search yields this:
https://www.reddit.com/r/rust/comments/6x4bhg/how_slow_is_arc_relative_to_rc/

Seems to have a non trivial impact, but a benchmark might be more informative. Not something to casually throw in though.

johanneskoester:
I think you are right. Better let the caller decide. I'll close this for now.

PR #235, opened 2020-08-16
Turn  the`bio-types` dependency into a feature

juliangehring:
Turns the dependency on `bio-types` into a feature. This allows opting-out of the dependency if integration with the `bio-types` is not needed. In addition, `bio-types` depends unconditonally on `serde` which can increase compile times significantly, and building it can't be avoided if `bio-types` is a required dependency.


pmarks:
@juliangehring I'm definitely a fan of using features to reduce dependencies (especially serde).  `bio-types` without `serde` is quite small, and we already have a `serde` feature in `htslib`.  So my suggestion for how to tackle this would be: add a `serde` feature to `bio-types`, and control it with the `serde` feature of `htslib`.

I think it could be quite confusing to not be able to access `strand()` by on a `Record` unless you turn on a certain feature.

pmarks:
@juliangehring want me to take a crack at a `serde` feature in bio-types, or are you already started on that?

juliangehring:
@pmarks Sure, feel free to go for it.

PR #239, opened 2020-08-20
(De)serializing of Cigar objects

FelixMoelder:
This PR derives serde serialization and deserialization for Cigar enum.
As it is necessary to import serde_derive feature via the serde crate I had to rename the `serde` feature defined in `Cargo.toml`.


Cargo.toml
```diff
@@ -25,6 +25,7 @@ regex = "1.3"
 linear-map = "1.2"
 serde_base = { version = "^1", optional = true, package = "serde" }
 serde_bytes = { version = "0.11", optional = true }
+serde = { version = "^1", features = ["derive"] }
```
pmarks, line 28: Don't add this line -- line 26 is already importing `serde` via `package = "serde"`

Cargo.toml
```diff
@@ -39,7 +40,7 @@ libdeflate = ["hts-sys/libdeflate"]
 s3 = ["hts-sys/s3"]
 gcs = ["hts-sys/gcs"]
 static = ["hts-sys/static"]
-serde = ["serde_base", "serde_bytes"]
+serde_features = ["serde_base", "serde_bytes"]
```
pmarks, line 43: If you revert the change on line 28, you be able to able revert this change.  There's no need to change the name of the feature. (`serde` is the standard name used in all crates that do this, and changing the feature name break backward compatibility).

src/bam/record.rs
```diff
@@ -867,7 +867,7 @@ impl<'a> ops::Index<usize> for Seq<'a> {
 unsafe impl<'a> Send for Seq<'a> {}
 unsafe impl<'a> Sync for Seq<'a> {}
 
-#[derive(PartialEq, Eq, Debug, Clone, Copy, Hash)]
+#[derive(PartialEq, PartialOrd, Eq, Debug, Clone, Copy, Hash, Deserialize, Serialize)]
```
pmarks, line 870: Yo need to add the serde derives in a separate line like this:
`#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]`

FelixMoelder: I've reverted most of the changes but this does not seem to work just yet, as `Serialize` and `Deserialize` are out of scope.
Instead I had to define `serde_derive` as feature.
While this is compiling and all tests are passing serializing a Cigar still does not work returning `the trait `cli::_::_serde::Serialize` is not implemented for `rust_htslib::bam::record::Cigar``

FelixMoelder:
> @FelixMoelder my apologies, we will need to rename the feature as you did originally. (The other option would be to require 'serde' and 'serde_bytes' to be switched on simultaneously, which is a bit more difficult I think). I reverted back to your original style & fixed the `cfg` blocks. Does this look right to you now?

This looks good! I've just tested your changes and everything is working fine, now. Thanks a lot.

johanneskoester (approved):

johanneskoester:
Thank you guys!

PR #241, opened 2020-08-24
shared thread pool

pmarks:


src/tpool.rs
```diff
@@ -0,0 +1,54 @@
+use std::cell::RefCell;
+use std::rc::Rc;
+
+pub use crate::bam::errors::{Error, Result};
+use crate::htslib;
+
+/// An HTSlib thread pool. Create a thread pool and use `set_thread_pool()` methods
+/// to share a thread pool across multiple BAM readers & writier.
```
nlhepler, line 8:
```suggestion
/// to share a thread pool across multiple BAM readers & writers.
```

src/tpool.rs
```diff
@@ -0,0 +1,54 @@
...
+/// The Rust wrapper holds the htslib thread pool behind a Rc, and a Rc reference
+/// to the thread pool is held by each reader / writer so you don't need to
+/// explicitly manage the lifetime of the `ThreadPool`.
+#[derive(Clone, Debug)]
+pub struct ThreadPool {
+    pub(crate) handle: Rc<RefCell<InnerThreadPool>>,
```
nlhepler, line 14: Should this be an Arc so you can share it across threads?

brainstorm: We recently had https://github.com/rust-bio/rust-htslib/pull/234, would be interesting to hear your opinions on that one (related to `Arc` discussion(s)).

brainstorm (approved):
Good stuff! I like the way @pmarks reuses htslib's threadpool instead of going for more complex options such as rayon & co on the Rust level. Simple to understand and use.

Other than the `ARc` vs `Rc` discussion, I'm all in with those changes :+1:

brainstorm:
Alright, merging this. At the end of the day it's better for perf to have a thread pool than no threads at all (leaving Rc/Arc discussion aside) ;)

PR #248, opened 2020-08-28
tweak name of "serde" feature -> "serde_feature" in the README

pjedge:
the name of the feature seems to have changed since the last version


pmarks:
> the name of the feature seems to have changed since the last version

that's my bad. Thanks for the PR~

PR #250, opened 2020-08-30
Fixes for the bcf submodule

holtgrewe:
The first patch unpacks BCF records when read through IndexedReadres. This is required, e.g., for `Record::id()` to properly work.

The second patch properly recognizes missing files in BCF readers. Previously, a UTF 8 path error was reported on missing files.


src/bcf/mod.rs
```diff
@@ -89,9 +89,15 @@ pub unsafe fn set_threads(hts_file: *mut htslib::htsFile, n_threads: usize) -> R
...
-            _ => Err(errors::Error::NonUnicodePath),
+        if !path.as_ref().exists() {
+            Err(Error::FileNotFound {
+                path: path.as_ref().to_owned(),
+            })
+        } else {
+            match path.as_ref().to_str() {
+                Some(p) if path.as_ref().exists() => Ok(Self::new(p.as_bytes())?),
+                _ => Err(errors::Error::NonUnicodePath),
+            }
```
tedil, line 100: I'd keep the match and just manage the arms differently, something like this:
```suggestion
        match path.as_ref().to_str() {
            Some(p) if !path.as_ref().exists() => Err(Error::FileNotFound { path: p.into() }),
            Some(p) => Ok(Self::new(&ffi::CString::new(p).unwrap())?),
            _ => Err(Error::NonUnicodePath),
        }
```

src/bcf/errors.rs
```diff
@@ -36,6 +38,8 @@ pub enum Error {
     DuplicateSampleNames,
     #[error("invalid (non-unique) characters in path")]
     NonUnicodePath,
+    #[error("file not found: {path}")]
+    FileNotFound { path: PathBuf },
```
tedil, line 42:
```suggestion
    #[error("file not found: {path}")]
    FileNotFound { path: String },
```

src/bcf/mod.rs
```diff
@@ -178,9 +184,15 @@ impl IndexedReader {
...
-            _ => Err(Error::NonUnicodePath),
+        if !path.as_ref().exists() {
+            Err(Error::FileNotFound {
+                path: path.as_ref().to_owned(),
+            })
+        } else {
+            match path.as_ref().to_str() {
+                Some(p) if path.as_ref().exists() => Ok(Self::new(&ffi::CString::new(p).unwrap())?),
+                _ => Err(Error::NonUnicodePath),
+            }
```
tedil, line 195: see above

tedil (changes requested):

brainstorm:
Ugh, sorry @tedil @holtgrewe I thought that `match` suggestion was a working/tested one before committing it.

src/bcf/mod.rs
holtgrewe: What exactly do you mean here?

tedil: Ah, sorry my bad; what I meant was that this can also be more concise using the `match` version I suggested for the IndexedReader.

tedil:
> Ugh, sorry @tedil @holtgrewe I thought that `match` suggestion was a working/tested one before committing it.

My bad, I should have checked it's correct before tagging it as a suggested change!

holtgrewe:
Maybe it would make sense to split the "always unpack" change from the other?

src/bcf/mod.rs
```diff
@@ -178,9 +179,15 @@ impl IndexedReader {
...
```
tedil, line 190: Does this work?
```suggestion
        let path = path.as_ref();
        match path.to_str() {
            Some(p) if path.exists() => Self::new(&ffi::CString::new(p)?),
            Some(p) => Err(Error::FileNotFound { path: p.into() }),
            None => Err(Error::NonUnicodePath),
        }
```

tedil (changes requested):

johanneskoester:
@holtgrewe, once the conflicts are resolved, I think we can merge an release, as @tedil has hopefully committed a fixed version of the match.

tedil:
> @holtgrewe, once the conflicts are resolved, I think we can merge an release, as @tedil has hopefully committed a fixed version of the match.

So… I hope I didn't mess up by merging changes from master into this fork/branch in order to resolve the conflicts, but at least all checks have passed.

johanneskoester (approved):

tedil (approved):

Issue #162, opened 2019-10-29
read api changes broke my code :(. / while let Ok(_) = read() no longer works


TyberiusPrime:
This is mostly a cautionary tale for others:

The alloc-less way to iterate over reads used to be based on 
```fn read(&mut self, record: &mut record::Record) -> Result<(), ReadError>```
which has recently (version 0.25) converted to
``` fn read(&mut self, record: &mut record::Record) -> Result<bool> ```

Accordingly, the calling code most be adjust from 
```while let Ok(_) = bam.read(&mut record)```
to 
```while let Ok(true) = bam.read(&mut record)```

Otherwise you run into an endless loop.

Unfortunatly, the Ok(_) pattern is valid rust even on the new types so no compiler errors.

This has been a fun hour of doubting my sanity.

**Comment by johanneskoester** [comment]
I am so sorry, I did not anticipate that side-effect.

**Comment by brainstorm** [comment]
Wouldn't it make sense to keep (or rollback to) `Result<(), ReadError>` in rust-htslib instead of `Result<bool>` which is not very Rust-idiomatic (common?) AFAIK?

I can [see other functions](https://github.com/rust-bio/rust-htslib/blob/master/src/bcf/mod.rs#L45) with this gotcha and it can indeed be a trap for young #rust-htslib players like me, IMHO :/

**Comment by sjackman** [comment]
Does it every return `Ok(false)`, and if so, what does that mean? If it never returns `Ok(false)` then I agree that this change would best be rolled back to instead return `Result<()>`.

**Comment by slazicoicr** [comment]
I am also not fond of the `Result<bool>` approach (user has to look up what true/false means), but neither do I think `Result<()>` was ideal (user has to look up which error means the end of file). I think the use of `Option` is the most explicit and clear representation: `Option<Result<()>>`. Am I getting any reads and if so, was there an error in parsing the read?

This would informative compile time errors for anyone who used the `while let Ok(_) = bam.read(&mut record)` pattern. It would require minor modification:

```
while let Some(res) = bam.read(&mut record) {
    // deal with res if not Ok, otherwise use record
}

brainstorm:
@sjackman I've not done a exhaustive checking on all functions and what they return, but the return type that @slazicoicr proposes makes sense to me.

[...]

PR #257, opened 2020-09-11
Change `read` API to Option<Result<()>>

slazicoicr:


brainstorm:
@dlaehnemann @slazicoicr @TyberiusPrime (and everyone) what shall we do about this PR and https://github.com/rust-bio/rust-htslib/issues/162 overall? :)

TyberiusPrime:
Since the recent hackaton is basically all breaking changes on "fetch" any way
(you have to change fetch(chr, start, stop) to fetch((chr, start, stop)), 
we might as well get this in the same release.

Option<Result<_> is what an iterator would return (internally) - I think it's appropriate.

Side note: 
I've mostly prepared one more PR that replaces all 'fetch, then iterate' APIs with 'fetch-and-iterate' ones (basically the missing part of #163),but it's not ready yet (and I'm sitting at the wrong machine this morning), and it's orthogonal to the return type of read().

dlaehnemann:
So I guess, we merge this and then at least wait for this last PR by @TyberiusPrime before a new release. And maybe check again, if there is more breaking stuff that could go in -- while we are at it...

dlaehnemann (approved):
LGTM

dlaehnemann:
I'd say this is ready to merge, but after I meddled so much, maybe someone else wants to quickly scan the code and merge? @brainstorm ;)

brainstorm:
Thanks much for this @slazicoicr !

PR #258, opened 2020-09-12
Add module to expose the bgzf interface from htslib into the crate.

landesfeind:
This change adds a `bgzf` module that wraps around the corresponding htslib functions. Currently, only a streamed reader is implemented.

There are a couple of alternative crates available:
- [bgzip](https://crates.io/crates/bgzip)
- [deflate](https://crates.io/crates/deflate)
- [flate2](https://crates.io/crates/flate2)

However, I still this it makes sense to expose the bgzf function in htslib because
- htslib (transparently) reads uncompressed, zip, and bgzipped files (it's nice for a developer to not care)
- it is very likely that a user of rust-htslib who wants to read BAM or VCF data also will need to associate with data from other text-files like GTF, BED, etc. It is very convenient to not having to include another crate if htslib already provides the functionality


pmarks (approved):
Cool, looks good to me.  One additional improvement would be to integrate this with the htslib thread pool. See PR #241 for details. You can use the bgzf_thread_pool method in the htslib API to make a given BGZF file handle use a given thread pool to decompress data, which provides parallel decompression of bgzf blocks.

landesfeind:
Thanks for the feedback and the idea - I really like :-)

dlaehnemann:
Yes, very nice addition. And another point in favour: the respective htslib functionality is being debugged and improved, actively. So no worries that this might negatively impact performance here, but rather the other way around: we'll directly profit from any improvements!

PR #260, opened 2020-09-13
Openssl version bump

brainstorm:


dlaehnemann (approved):
I am assuming this is just to increase security, right? Looks good to me, as it simply increases lower boundary of admissible crate version to the latest.

brainstorm:
> I am assuming this is just to increase security, right? Looks good to me, as it simply increases lower boundary of admissible crate version to the latest.

That's exactly right, my thoughts :)

PR #261, opened 2020-09-14
Introduce ergonomic fetch(enum) api (half of #163)

TyberiusPrime:
I extracted, rewrote and extended the proposed fetch API changes from #163.

Drawback: it's a breaking change for fetch(tid, start, stop) and fetch_str("chr:start-stop"), 
which have to be replaced by fetch((tid, start, stop) and fetch("chr:start-stop").

Advantages: fetch now takes anything remotely sensible:
* (tid-or-reference-name, start, stop)
* (tid-or-reference-name)
* "chr:start-stop"
with tids being accepted as i32 or u32 (with panic if too large), reference names can be &[u8], or &str, start stop may be any of i32/u32/i64/u64 (with panics if exceeding i64).

The latter is necessary since our header functions are confused after the 'large positions' upgrade in htslib, and for example header.target_len returns an u64. This keeps everything ergonomic until we have that fixed.


src/bam/mod.rs
```diff
@@ -349,6 +349,153 @@ impl Drop for Reader {
...
+    }
+}
+
+impl<'a> From<&'a str> for FetchDefinition<'a> {
+    fn from(s: &'a str) -> FetchDefinition<'a> {
+        FetchDefinition::String(s.as_bytes())
+    }
+}
+
+//also accept &[u8;n] literars
```
tedil, line 454:
```suggestion
// also accept &[u8;n] literals
```

src/bam/mod.rs
```diff
@@ -434,7 +581,68 @@ impl IndexedReader {
...
+    /// * FetchDefinition::All or "." -> Fetch overything
+    /// * FetchDefinition::Unmapped "*" -> Fetch unmapped (as signified by the 'unmapped' flag in the BAM - might be
+    ///   unreliable with some aligners.
+    ///
+    /// This replaces the old fetch and fetch_str implementations.
+    pub fn fetch<'a, T: Into<FetchDefinition<'a>>>(&mut self, fetch_definition: T) -> Result<()> {
+        //this 'compile time redirect' safes us
+        //from monomorphing the 'meat' of the fetch function
+        self._inner_fetch(fetch_definition.into())
+    }
```
tedil, line 602: This doc comment should have at least one usage example I think! Or even multiple examples to showcase the different variants

tedil (changes requested):
Looks pretty good to me, but I'd really like to have some examples in the doc comments for `fetch`!

src/bam/mod.rs
TyberiusPrime: Added. The example-boilerplate to actual fetch calls ratio doesn't warrant multiples though.

src/bam/mod.rs
```diff
@@ -434,7 +581,85 @@ impl IndexedReader {
...
+    ///
+    /// Both iterating (with [.records()](trait.Read.html#tymethod.records)) and looping without allocation (with [.read()](trait.Read.html#tymethod.read) are a two stage process:
+    /// 1. 'fetch' the region of interest
+    /// 2. iter/loop trough the reads.
+    ///
+    /// Example:
+    /// ```
+    /// use rust_htslib::bam::{IndexedReader, Read};
+    /// let mut bam = IndexedReader::from_path(&"test/test.bam").unwrap();
+    /// bam.fetch(("chrX", 10000, 20000)); // coordinate 10000..20000 on reference named "chromosomeX"
```
tedil, line 594:
```
    /// bam.fetch(("chrX", 10000, 20000)); // coordinates 10000..20000 on reference named "chrX"
```

brainstorm (approved):
Really nice indeed, thanks @TyberiusPrime 👍🏻

tedil (approved):

src/lib.rs
```diff
@@ -102,6 +102,12 @@
 //! bam.fetch(("CHROMOSOME_I", 0, 20)).unwrap();
 //! // afterwards, read or pileup in this region
 //! ```
+//!
+//! See
```
tedil, line 106:
```suggestion
//! See
```

PR #262, opened 2020-09-14
Introduced BamRecordExtensions::aligned_block_pairs

TyberiusPrime:
You could get read&genome coordinates base-by-base,
or you could get genome-coordinates in continuously aligned blocks,
but you couldn't get read&genome coordinates in blocks.

Since I recently ran into a situtation where not allocating every position just to iterate through them shaved of a few percentages of my runtime, 
I thought I might as well contribute the combined function.

### DIFF
--- a/src/bam/ext.rs
+++ b/src/bam/ext.rs
@@ -48,7 +48,6 @@ impl Iterator for IterAlignedBlockPairs {
     }
 }
 
-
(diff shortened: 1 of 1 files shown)


TyberiusPrime:
I have also an iterator implementation at hand, if #267 were to be approved.

brainstorm (approved):

Issue #228, opened 2020-07-28
Count alt-allele dosage in multi-sample VCF/BCF


weinstockj:
Hi Rust-htslib developers, 

I'm looking for the best way to get the alt-allele dosage for a given variant in a multi-sample vcf/bcf. For example, for a bi-allelic SNP, I'd like to convert the vector of genotypes to a vector of 0 (Hom-ref), 1 (Het), or 2 (Hom-alt). Phase is ignored in this calculation. 

At the moment, I am doing something like (which counts alt-alleles and stores in a Hashmap):

```
 let genotypes = record.genotypes().unwrap();
 for s in 0..sample_count {
                let mut alt = 0;
                if genotypes.get(s)[0] == bcf::record::GenotypeAllele::Unphased(1) {
                    alt += 1;
                }
                if genotypes.get(s)[1] == bcf::record::GenotypeAllele::Unphased(1) {
                    alt += 1;
                }

                if map.contains_key(&samples[s]) {
                    *map.entry(&samples[s]).or_insert(0) += alt;
                } else {
                    map.insert(&samples[s], alt);
                }
  }

```
Storing these as a hashmap vs vector is just a detail. Curious if this it the recommended approach for working with unphased genotypes in a vcf/bcf. 

Thanks for developing rust-htslib,
Josh

juliangehring:
I'm not sure if there is a recommended or best way of achieving this. Here is one alternative suggestion - maybe this is helpful if you also want to handle non-bi-allelic variants:

```rust
let gts = record.genotypes().unwrap();
for sample_index in 0..sample_count {
    // for each sample
    for gta in gts.get(sample_index).iter() {
        // for each allele
        match gta.index() {
            Some(0) => n_ref[sample_index] += 1, // reference allele
            Some(_) => n_alt[sample_index] += 1, // alt allele
            None => break,                       // missing allele
        }
    }
}
```

weinstockj:
got it - that is helpful. Thanks!

brainstorm:
Closing in favor of #198 ;)

brainstorm:
Reopening, since this issue was very specific and well written/formulated, worth adding to the official docs, thanks both!

brainstorm:
@edmundlth working on this 💪🏻

PR #263, opened 2020-09-14
Add testable examples to doc strings for bcf::{Record, HeaderView}

edmundlth:
This includes the example in https://github.com/rust-bio/rust-htslib/issues/228. 
The idea is that these examples will show up in a searchable manner in the official doc and `cross test` will execute them for unit testing.


src/bcf/header.rs
```diff
@@ -236,6 +236,33 @@ pub struct HeaderView {
...
+/// }
+/// assert_eq!(header.sample_id(b"one").unwrap(), 0);  // Sample index wrapped in Option<usize>
+/// assert_eq!(header.sample_id(b"two").unwrap(), 1);  // Sample index wrapped in Option<usize>
+/// assert!(header.sample_id(b"non existent sample").is_none());  // Return none if not found
+///
+/// assert_eq!(header.contig_count(), 1); // Number of contig in header.
+/// // obtain the data type of an INFO field
+/// let (tag_type, tag_length) = header.info_type(b"S1").unwrap();
+/// let (fmt_type, fmt_length) = header.format_type(b"GT").unwrap();
+/// ```
```
dlaehnemann, line 265: I really like these explanatory doctests!

And I think the optimum where to put tests in this case is a bit complicate and this might not be the right place for them. I would try to do two things:
1. Split this bigger test up a bit and have small little tests below for each of the functions defined for a `HeaderView`, so for `sample_count()`, `samples()`, `sample_id()`, etc. This might seem a bit repetative, but if in doubt you can [keep repetitive code from showing up](https://doc.rust-lang.org/rustdoc/documentation-tests.html#hiding-portions-of-the-example). And where possible, you can even explain input parameters, [this page gives a nice and quick intro to function-specific doc comments](https://doc.rust-lang.org/stable/rust-by-example/meta/doc.html).
2. Move whatever is useful of this example code up to the `bcf::header` module level. So move it to the top of `src/bcf/header.rs` (after imports) and use the module level `//!` doc comment syntax ([see here for possible types of doc comments](https://doc.rust-lang.org/stable/reference/comments.html#doc-comments)). This way, they [will show up in the docs here](https://docs.rs/rust-htslib/0.32.0/rust_htslib/bcf/header/index.html), saving users from having to click on `HeaderView` or `Header` to get usable code examples and indicating on which to click for further infos on the exact functions.

If anything is unclear regarding those functions, please ask away here or over at Discord!

src/bcf/mod.rs
```diff
@@ -1296,4 +1298,32 @@ mod tests {
...
+                    // for each allele
+                    match gta.index() {
+                        Some(0) => n_ref[sample_index] += 1, // reference allele
+                        Some(_) => n_alt[sample_index] += 1, // alt allele
+                        None => break,                       // missing allele
+                    }
+                }
+            }
+        }
+    }
```
dlaehnemann, line 1328: Great idea adding this as a test case, as well. But this should then probably include some [`assert_eq!()`](https://doc.rust-lang.org/std/macro.assert_eq.html) invocations at the end to check that the right counts were generated?

src/bcf/record.rs
```diff
@@ -74,6 +74,45 @@ impl NumericUtils for i32 {
...
+///             // for each allele
+///             match gta.index() {
+///                 Some(0) => n_ref[sample_index] += 1, // reference allele
+///                 Some(_) => n_alt[sample_index] += 1, // alt allele
+///                 None => break,                       // missing allele
+///             }
+///         }
+///     }
+/// }
+/// ```
```
dlaehnemann, line 115: I think, this best goes into the main file of the module (`src/bcf/mod.rs`) as module-level documentation (`//!`) that will eventually show up in the docs right here: https://docs.rs/rust-htslib/0.32.0/rust_htslib/bcf/index.html

src/bcf/record.rs
```diff
@@ -357,6 +396,18 @@ impl Record {
...
+    /// ```
+    /// use crate::rust_htslib::bcf::{Reader, Read};
+    /// let mut vcf = Reader::from_path(&"test/test_string.vcf").expect("Error opening file.");
+    /// let expected = ["./1", "1|1", "0/1", "0|1", "1|.", "1/1"];
+    /// for (rec, exp_gt) in vcf.records().zip(expected.iter()) {
+    ///     let mut rec = rec.expect("Error reading record.");
+    ///     let genotypes = rec.genotypes().expect("Error reading genotypes");
+    ///     assert_eq!(&format!("{}", genotypes.get(0)), exp_gt);
+    /// }
+    /// ```
```
dlaehnemann, line 410: Great function-level doctest!

dlaehnemann (commented):
Thanks for tackling this. I have some recommendations for moving tests to other places and some more pointers to rust doctest documentation. Please feel free to doubt any of my suggestions and ask away!

edmundlth:
Sorry for the delay.. I have move those larger usage examples to module level documentation. When I get the time, I will get back into writing function level docs.

johanneskoester (approved):

johanneskoester:
Thanks a lot!

PR #266, opened 2020-09-14
Explicitly reference std::i64::MAX instead of i64::MAX

landesfeind:
Compiling with older versions of rust like v1.41.0 failed because it does not directly resolve i64 into `std`. To reproduce run
```bash
rustup run 1.41.0 cargo build
``` 
which results in error:
```
Compiling rust-htslib v0.32.1-alpha.0 (/data/rust/rust-htslib)
error[E0599]: no associated item named `MAX` found for type `i64` in the current scope
  --> src/faidx/mod.rs:78:25
   |
78 |         if begin > i64::MAX as usize {
   |                         ^^^ associated item not found in `i64`
   |
help: you are looking for the module in `std`, not the primitive type
   |
78 |         if begin > std::i64::MAX as usize {
   |                    ^^^^^^^^^^^^^

error[E0599]: no associated item named `MAX` found for type `i64` in the current scope
  --> src/faidx/mod.rs:81:23
   |
81 |         if end > i64::MAX as usize {
   |                       ^^^ associated item not found in `i64`
   |
help: you are looking for the module in `std`, not the primitive type
   |
81 |         if end > std::i64::MAX as usize {
   |                  ^^^^^^^^^^^^^

### DIFF
--- a/src/faidx/mod.rs
+++ b/src/faidx/mod.rs
@@ -75,10 +75,10 @@ impl Reader {
     /// * `begin` - the offset within the template sequence (starting with 0)
     /// * `end` - the end position to return (if smaller than `begin`, the behavior is undefined).
     pub fn fetch_seq<N: AsRef<str>>(&self, name: N, begin: usize, end: usize) -> Result<&[u8]> {
-        if begin > i64::MAX as usize {
+        if begin > std::i64::MAX as usize {
             return Err(Error::PositionTooLarge);
         }
-        if end > i64::MAX as usize {
+        if end > std::i64::MAX as usize {
             return Err(Error::PositionTooLarge);
         }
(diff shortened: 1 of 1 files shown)


TyberiusPrime:
Do we have a 'Minimum supported Rust version policy'?

dlaehnemann (approved):
I am not aware of any minimum version of Rust that we (aim to) support. But this seems straightforward to simply do...

TyberiusPrime:
Not complaining about this PR, just wondering if we should have one.

dlaehnemann:
Yeah, no worries. Not sure we need this, though. My feeling would rather be, that tool development either moves rather quickly alongside `rust-htslib` development or maintainers can freeze their tool with a certain version of `rust-htslib`...

PR #267, opened 2020-09-16
BamRecordExtensions now return Iterators instead of Vec<>

TyberiusPrime:
This turns our BamRecordExtensions into iterators instead of allocating vectors.
Should be more flexible, and possibly a bit faster since it safes on allocations.

I should've done it this way in the first place, but was lacking the Rust foo back then.


brainstorm (approved):
Comprehensive, better than alloc as discussed, love it!

brainstorm (approved):

PR #268, opened 2020-09-16
fixed clippy warnings

TyberiusPrime:
We had a handful of clippy warnings, now we don't.

This also 'un-pubs' two unsafe functions that are used between Reader classes - not sure why these were public in the first place?


pmarks (approved):
LGTM! thanks!

Issue #269, opened 2020-09-21
Behaviour with S3 cross region is different between Rust binding lib and regular htslib


andrewpatto:
Using htsfile to do a basic AWS S3 access of a file in us-west-2. This works.

`AWS_PROFILE=sandbox htsfile -vvvvvvvvvv -h s3://igv-demo-bucket-west/TD01-GV1001.merged.bam`

we can see in the logs the initial connection to us-east-1 that generates a 400 error, which is then parsed for redirect and corrected..

```
[D::init_add_plugin] Loaded "knetfile"
[D::init_add_plugin] Loaded "mem"
[D::init_add_plugin] Loaded "libcurl"
[D::init_add_plugin] Loaded "gcs"
[D::init_add_plugin] Loaded "s3"
[D::init_add_plugin] Loaded "s3w"
*   Trying 52.218.219.51...
* TCP_NODELAY set
* Connected to igv-demo-bucket-west.s3.amazonaws.com (52.218.219.51) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/cert.pem
  CApath: none
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; ST=Washington; L=Seattle; O=Amazon.com, Inc.; CN=*.s3.amazonaws.com
*  start date: Nov  9 00:00:00 2019 GMT
*  expire date: Mar 12 12:00:00 2021 GMT
*  subjectAltName: host "igv-demo-bucket-west.s3.amazonaws.com" matched cert's "*.s3.amazonaws.com"
*  issuer: C=US; O=DigiCert Inc; OU=www.digicert.com; CN=DigiCert Baltimore CA-2 G2
*  SSL certificate verify ok.
> GET /TD01-GV1001.merged.bam HTTP/1.1
Host: igv-demo-bucket-west.s3.amazonaws.com
User-Agent: htslib/1.10.2 libcurl/7.64.1
Accept: */*
Authorization: AWS4-HMAC-SHA256 Credential=AKIAZWW**********/20200921/us-east-1/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date,Signature=d92309656430279bea7a12c2f4f0913ef0ed8eeee7e6e4999d7c96646d4816bc
[...]

PR #270, opened 2020-09-23
Htslib 1.11

brainstorm:
Fixes issue #269.


brainstorm:
@pmarks While you are over here (#273), would you mind taking a peek at this one and approve if it looks good, pleeease? ;)

pmarks (approved):
LGTM - I don't see anything in release notes that would require us to change anything.

PR #273, opened 2020-09-24
fix detection of stale htslib files

### LINKED ISSUE #271: Inconsistent build behaviour for some .c files in hts-sys/htslib
This came about from some of my s3 bug hunting - I was editing `hts-sys/htslib/hfile_s3.c` to add some debug traces - all was fine in that whenever I made changes to that file a `cargo build` would cause a C compile. (btw this was being used from a project that selected the `s3` feature).

As I dug deeper at some point I needed to add some debug traces to `hts-sys/htslib/hfile.c` - which to my surprise *does not* trigger any rebuild after editing - but only when the `s3` feature is enabled.

I took a brief look at the `build.rs` and `hfile.c` is definitely listed as `cargo:rerun-if-changed=` - so it feels like it should work - but I am certainly not an expert in rust build systems. 

(the shortest path to recreate for me is - just tested (macos))

- checkout rust-htslib
- `cargo build --features s3`
- `vi hts-sys/htslib/hfile.c` (add syntax error at top of file)
- `cargo build --features s3` (this will succeed when it shouldn't)
- `cargo build` (this will fail with a syntax error)
- `vi hts-sys/htslib/hfile_s3.c` (add syntax error at top of file)
- `cargo build --features s3` (this will fail as expected)

So it appears the s3 feature disables dependency checking for `hfile.c` - though its not like `hfile.c` is an optional feature - it is the core file for making all the other file reader plugins work.

@brainstorm suggested I ping @nlhepler and @pmarks as experts in the build.rs..

pmarks:
fixes #271.  We make a fresh copy of the `htslib` submodule everytime `build.rs` runs. this is a bit slower than strictly needed, but is protected by the `rerun-if-changed` checks on the main copy of the htslib files.

### DIFF
--- a/hts-sys/build.rs
+++ b/hts-sys/build.rs
@@ -60,13 +60,15 @@ const FILES: &[&str] = &[
 fn main() {
 
     let out = PathBuf::from(env::var("OUT_DIR").unwrap());
-    if !out.join("htslib").exists() {
-        println!("copying...");
-        copy_directory("htslib", &out).unwrap();
+    let htslib_copy = out.join("htslib");
+
+    if htslib_copy.exists() {
+        std::fs::remove_dir_all(htslib_copy).unwrap();
     }
+    copy_directory("htslib", &out).unwrap();
     
     let mut cfg = cc::Build::new();
-    
+
     // default files
     let out_htslib = out.join("htslib");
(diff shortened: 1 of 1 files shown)


brainstorm (approved):
Good old wipe, LGTM ;)

Issue #224, opened 2020-07-08
Support for being able to add genotypes to a BCF record


hjeremyli:
As far as I understand, there is currently no way to update genotypes in a bcf `Record`.

Related:
https://github.com/rust-bio/rust-htslib/blob/ed217c2fe460794d249e58a85db0b0c8a2929e55/src/bcf/record.rs#L360

Since the `GT` field has a `String` format (as specified in the header) I tried to work around this by using `Record.push_format_string()` like so:

```rust
fn main() {
    let mut header = rust_htslib::bcf::header::Header::new();
     header.push_record(format!("##FORMAT=<ID=GT,Number=1,Type=String,Description=\"Genotype\">").as_bytes());
    header.push_sample("samplename".as_bytes());
    // ... add some other stuff to the header...

    let mut writer = rust_htslib::bcf::Writer::from_path("out.vcf.gz", &header, false, rust_htslib::bcf::Format::VCF).unwrap();
    let mut rec = writer.empty_record();    

    // ... populate chr, pos, id, etc. .... 

    // attempt to add genotype field using fixed string 
    let gt_test = vec!["0/0".as_bytes()];
    rec.push_format_string("GT".as_bytes(), &gt_test).unwrap();

    writer.write(&rec).unwrap();
}
```

However, this yields the following error:

```[E::bcf_format_gt] Unexpected type 7
```

This above works if I replace the tag argument in `push_format_string()` with something other than `"GT"`, which seem to suggest that attempting to edit the GT field using this method is treated differently than for other methods. 

It seems I'm likely missing something here, but is there a straightforward way of updating genotypes for a given record as I attempted to do above with the current release? Something I imagine that would provide similar functionality to the `bcf_update_genotypes` macro in `htslib`.

MaltheSR:
I needed to do the same thing and played around with this for a bit. In case anyone else stumbles on this, the GT FORMAT field is apparently special in that it is marked as a string in the header, but it is actually encoded as an integer. (I found [this](https://github.com/samtools/htslib/issues/54) discussion helpful on that point.) 

Therefore, corresponding to the example above, I've been able to do e.g.:

```
let gt_test: [i32; 2] = [2, 2];
record.push_format_integer("GT".as_bytes(), &gt_test).unwrap();
```

[...]

PR #277, opened 2020-09-30
Helper to push genotypes to BCF record

MaltheSR:
With reference to [#224 ](https://github.com/rust-bio/rust-htslib/issues/224), pull request includes:

1. Write `impl From<GenotypeAllele> for i32 {}` to encode `GenotypeAllele`s according to BCF specs.
2. Write `push_genotypes` helper for Record to add an array of `GenotypeAllele`s to `Record`, which is a simple wrapper around `GenotypeAllele` conversion and `push_format_integer`.

I've also added a bit of testing for both. I also mean to try and improve the documentation for pushing information onto records (as mentioned in the issue), but I thought to keep that in a later pull request once this is settled.

I'm still getting used to Rust and pull requests; feedback very welcome.

### DIFF
--- a/src/bcf/record.rs
+++ b/src/bcf/record.rs
@@ -365,10 +365,7 @@ impl Record {
     ///
     /// Returns error if GT tag is not present in header.
     pub fn push_genotypes(&mut self, genotypes: &[GenotypeAllele]) -> Result<()> {
-        let encoded: Vec<i32> = genotypes
-            .iter()
-            .map(|gt| i32::from(*gt))
-            .collect();
+        let encoded: Vec<i32> = genotypes.iter().map(|gt| i32::from(*gt)).collect();
         self.push_format_integer("GT".as_bytes(), &encoded)
     }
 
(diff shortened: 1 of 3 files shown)


src/bcf/record.rs
```diff
@@ -354,7 +354,20 @@ impl Record {
...
+    ///
+    /// - `genotypes` - a flattened, two-dimensional array of GenotypeAllele,
+    ///                 the first dimension contains one array for each sample.
+    ///
+    /// # Errors
+    ///
+    /// Returns error if GT tag is not present in header.
+    pub fn push_genotypes(&mut self, genotypes: &[GenotypeAllele]) -> Result<()> {
+        let encoded: Vec<i32> = genotypes.iter().map(|gt| i32::from(*gt)).collect();
+        self.push_format_integer("GT".as_bytes(), &encoded)
```
tedil, line 369: Not really important, but we might as well use the bytes literal here:
```suggestion
        self.push_format_integer(b"GT", &encoded)
```

tedil (approved):
This looks pretty good to me, thanks for your effort! :)

brainstorm:
Thanks @MaltheSR !  :)

PR #285, opened 2020-12-10
GenotypeAllele encoding and documentation for pushing to Record

MaltheSR:
Addresses the remaining points in #224: 

- Adds `From<i32> for GenotypeAllele`, copying the code from the old `from_encoded` method. Also included a bit of testing for this in the tests for the previous `From<GenotypeAllele> for i32`, ensuring that converting back and forth does nothing.
- Replaces the only internal usage of `from_encoded`, deprecating the method from `0.36.0` onwards.
- Adds module level documentation for setting up a minimal VCF writer with header and a record with chrom, pos, and genotype information.
- Adds illustrative examples/doctests for three write methods for `Record`: `set_rid`, `push_genotypes`, and `push_format_float`.

The doc tests all require quite a bit of setup to compile. I've hidden this from the doc, with a note about the presuppositions and a reference to the module level docs, which gives a full example. I'm not sure this is the best way to do it, but including all the setup code seems to me to divert attention from the main point. 

Moreover, the setup requires a lot of code duplication, which is not nice, but I couldn't find a way to centralise the setup so long as `#[cfg(doctest)]` doesn't work for this case (see Rust issue [#67295](https://github.com/rust-lang/rust/issues/67295)).

Very happy to hear any other takes on how to organise this.

### DIFF
--- a/src/bcf/mod.rs
+++ b/src/bcf/mod.rs
@@ -878,7 +878,7 @@ mod tests {
             // the artificial "not observed" allele is present in each record.
             assert_eq!(record.alleles().iter().last().unwrap(), b"<X>");
 
-            let mut fmt = record.format(b"PL");
+            let fmt = record.format(b"PL");
             let pl = fmt.integer().expect("Error reading format.");
             assert_eq!(pl.len(), 1);
             if i == 59 {
@@ -971,7 +971,7 @@ mod tests {
         let mut buffer = Buffer::new();
         for (i, rec) in vcf.records().enumerate() {
             println!("record {}", i);
-            let mut record = rec.expect("Error reading record.");
+            let record = rec.expect("Error reading record.");
             assert_eq!(
                 record
                     .info_shared_buffer(b"S1", &mut buffer)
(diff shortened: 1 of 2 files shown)


brainstorm (changes requested):
Looks good overall, thanks! As a first quick pass, I'd fix the clippy `variable does not need to be mutable` warnings, there are several that seem easy to fix?

MaltheSR:
I saw those, but they were from before this pull request, so I figured maybe they served some point and left them. Anyway, that should be fixed now. There's still a couple of clippy warnings relating to code I'm not familiar with, though.

brainstorm (approved):
Thanks @MaltheSR, I didn't realize the clippy warnings were there before, thanks for fixing!

I don't have strong opinions about your choices on the docs side of things, I reckon I'll defer  the verdict to the other two reviewers @tedil and @dlaehnemann ;)

dlaehnemann (approved):
I'll just add my two cents regarding the doctests: I really like this split of hiding complexity that is not relevant to the functionality that the doctest is actually testing and showing, but pointing to the module docs for the necessary setup. This sounds like the solution we've been looking for all along! :D

One tiny thing: You included some relative links within the docs. Is there a way to link to the doctest in the module docs, directly. That way, the hidden complexity would simply be one click away. But never mind if this is not quickly done.

tedil (approved):

MaltheSR:
Good idea, @dlaehnemann, and thank you for the reviews from all three of you. I've added direct links from the three doc tests that mention the module level docs. (I also fixed the monospace font on two other links I'd missed last time).

PR #289, opened 2021-01-06
add read orientation detection

johanneskoester:


src/bam/read_orientation.rs
```diff
@@ -0,0 +1,84 @@
...
+    R2R1,
+    None,
+}
+
+impl<'a> From<&'a bam::Record> for ReadOrientation {
+    /// Infer read orientation from given BAM record. Returns `ReadOrientation::None` if record
+    /// is not properly paired, mates are not mapping to the same contig, or mates start at the
+    /// same position.
+    fn from(record: &bam::Record) -> Self {
+        if record.is_paired() && record.is_proper_pair() && record.tid() == record.mtid() {
```
veldsla, line 42: Why exclude non proper pairs? Large insert sizes or unexpected orientation are often flagged as non-proper. This does not mean they have no orientation. It can also point to structural events. Requiring the same tid makes sense.

src/bam/read_orientation.rs
```diff
@@ -0,0 +1,84 @@
...
+            if record.pos() == record.mpos() {
```
veldsla, line 43: Why? Any adapter trimmed read has identical coordinates, but you can still infer the orientation?

src/bam/read_orientation.rs
```diff
@@ -0,0 +1,84 @@
...
+    R1F2,
+    R2F1,
+    F1F2,
+    R1R2,
+    F2F1,
+    R2R1,
+    None,
+}
+
+impl<'a> From<&'a bam::Record> for ReadOrientation {
```
veldsla, line 37: If the only way to get this orientation is using the from method, why is it in rust-htslib?

veldsla (commented):
Not sure why you would want this in rust-htslib. Added some comments inline.

src/bam/read_orientation.rs
johanneskoester: I see read orientation as a property of the mapped record. Hence, although it has to be inferred from the bam, it made sense to me to have it in here. We could have it in a separate crate, but that would need to be an entirely new one. It cannot be in rust-bio (because then rust-bio would need to depend on rust-htslib). It also cannot be in rust-bio-types, because rust-htslib depends on rust-bio-types. I think it is a reasonable convenience functionality that can be safely added here without bloating the codebase too much.

src/bam/read_orientation.rs
johanneskoester: Good point, I missed the long distance aspect. I'd rather change it to `!record.is_unmapped()` and `!record.is_mate_unmapped`

src/bam/read_orientation.rs
johanneskoester: The rationale was: if both reads start at the same position, there is no read that comes "first". How would you infer which one comes first in such a case?

src/bam/read_orientation.rs
veldsla: Sure, I get that, but then I would expect  a method on bam::Record itself. Also using the result of this function will probably involve another match statement which might not be the most optimal approach.

src/bam/read_orientation.rs
veldsla: Ah, I assumed the first_in_template was responsible for that (I didn't look very closely at the match table logic)

src/bam/read_orientation.rs
johanneskoester: Thinking about it, we could move `ReadOrientation` to bio-types, and implement instead `Into<ReadOrientation>` here in rust-htslib.

veldsla: That sounds reasonable. Should it be called ReadPairOrientation? The read orientation is represented by the flag is_reverse.

johanneskoester: Very good idea as well. I'll do that.

PR #297, opened 2021-03-09
Remove unused dependencies (deferable, strum_macros)

sreenathkrishnan:
This will reduce the number of crates pulled in when using `rust-htslib` crate


sreenathkrishnan:
@johanneskoester @pmarks A trivial chainge but helps cut down the number of crates pulled in when building `rust-htslib`

pmarks (approved):
nice!

Issue #299, opened 2021-04-25
Slow regex(?) parsing in CigarString::try_from()


mrvollger:
Hi,

I have been trying to learn rust by playing around with bioinformatics formats and rust-htslib has been great for that, so thanks!

I needed a function to turn a string slice into a cigar string from a PAF file for a whole genome alignment and I used [CigarString::try_from(&str)](https://docs.rs/rust-htslib/0.36.0/src/rust_htslib/bam/record.rs.html#1124-1168) to do it. I found that it took over a minute to parse a cigar string with ~200k elements which seemed slow. 

I have a feeling that maybe the regex crate is slowing this down because I cannot see anything else in the function that could be slowing it down. 

I rewrote the function without the regex and now it is much faster < 1s. But I am not a rust expert and my tests are not extensive so I didn't feel justified in making a pull request. But here is my code if it is of interest. 

Best,
Mitchell

```rust
/// Create a CigarString from given str.
/// # Example 
/// ```
/// use rustybam::paf;
/// use rust_htslib::bam::record::*;
/// use rust_htslib::bam::record::CigarString;
/// use rust_htslib::bam::record::Cigar::*;
/// use std::convert::TryFrom;
/// use std::str::FromStr;
/// let cigars = vec!["10M4D100I1102=", "100000M20=5P10X4M"];
/// for cigar_str in cigars{
///     let my_parse = paf::cigar_from_str(cigar_str).expect("Unable to parse cigar");
///     let hts_parse = CigarString::try_from(cigar_str).expect("Unable to parse cigar");
///     assert_eq!(my_parse, hts_parse);
/// }
/// ```
/// TODO add IO errors 
pub fn cigar_from_str(text: &str) -> Result<CigarString, std::io::Error> {
    let bytes = text.as_bytes();
    let mut inner = Vec::new();
    let mut i = 0;
    let text_len = text.len();
    while i < text_len {
        let mut j = i;
        while j < text_len && bytes[j].is_ascii_digit(){
            j += 1;
        }
        let n = u32::from_str(&text[i..j]).expect("Bam parse error");
        let op = &text[j..j+1];
        inner.push(match op {
            "M" => Cigar::Match(n),
            "I" => Cigar::Ins(n),
            "D" => Cigar::Del(n),
            "N" => Cigar::RefSkip(n),
            "H" => Cigar::HardClip(n),
            "S" => Cigar::SoftClip(n),
            "P" => Cigar::Pad(n),
[...]

PR #301, opened 2021-04-28
Potential speed up for CigarString::try_from(&str)

mrvollger:
This is a pull request based on: https://github.com/rust-bio/rust-htslib/issues/299#issuecomment-827160458

I needed a function to turn a string slice into a cigar string from a PAF file for a whole genome alignment and I used `CigarString::try_from(&str)` to do it, but I found that it took over a minute to parse a cigar string with ~200k elements which seemed slow.

I have a feeling that maybe the regex crate is slowing this down because I cannot see anything else in the function that could be slowing it down. I rewrote the function without the regex and now it is much faster < 1s. 

I moved the old function into the test module as `regex_cigar_from_str` and then wrote a new test (`test_cigar_parsing_vs_regex`) that compares the output of `regex_cigar_from_str` and the new implementation of `try_from`. 

I also added a check that requires Hard clipping only appear at the start or end of the cigar string as requested.

I ran `cargo test`, `cargo fmt`, and `cargo clippy`. `clippy` had some warnings but nothing new so I think it should be okay. But I am brand new to rust so I hope I didn't make any errors and I appreciate any comments/corrections! Thanks. 

-Mitchell

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -1159,7 +1159,19 @@ impl TryFrom<&str> for CigarString {
                         });
                     }
                 }
-                "S" => Cigar::SoftClip(n),
+                "S" => {
+                    if i == 0
+                        || j+1 == text_len
+                        || &text[i-1..i] == "H"
+                        || &text[j+1..].chars().all(|c| c.is_ascii_digit() || c == "H") {
+                        Cigar::SoftClip(n)
+                    } else {
+                        return Err(Error::BamParseCigar {
+                        msg: "Soft clips ('S') can only have hard clips ('H') between them and the end of the CIGAR string."
+                            .to_owned(),
+                        });
+                    }
+                },
                 "P" => Cigar::Pad(n),
                 "=" => Cigar::Equal(n),
                 "X" => Cigar::Diff(n),
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -1140,7 +1140,7 @@ impl TryFrom<&str> for CigarString {
             }
             // get the length of the operation
             let n = u32::from_str(&text[i..j]).map_err(|_| Error::BamParseCigar {
-                msg: "Unable to parse to u32".to_owned(),
+                msg: format!("Unable to parse &str '{}' to u32.", &test[i..j]).to_owned(),
             })?;
             // get the operation
             let op = &text[j..j + 1];
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -1170,7 +1170,7 @@ impl TryFrom<&[u8]> for CigarString {
                 }
                 b'S' => {
                     if i == 0
-                        || j+1 == text_len
+                        || j + 1 == text_len
                         || bytes[i-1] == b'H'
                         || bytes[j+1..].iter().all(|c| c.is_ascii_digit() || *c == b'H') {
                         Cigar::SoftClip(n)
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -1140,7 +1140,7 @@ impl TryFrom<&str> for CigarString {
             }
             // get the length of the operation
             let n = u32::from_str(&text[i..j]).map_err(|_| Error::BamParseCigar {
-                msg: format!("Unable to parse &str '{}' to u32.", &text[i..j]).to_owned(),
+                msg: format!("Unable to parse &str '{}' to u32.", &text[i..j]),
             })?;
             // get the operation
             let op = &text[j..j + 1];
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -1142,7 +1142,6 @@ impl TryFrom<&str> for CigarString {
     /// assert_eq!(cigar, expected_cigar);
     /// ```
     fn try_from(bytes: &[u8]) -> Result<Self> {
-        let bytes = text.as_bytes();
         let mut inner = Vec::new();
         let mut i = 0;
         let text_len = text.len();
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -1110,17 +1110,6 @@ impl CigarString {
 impl TryFrom<&[u8]> for CigarString {
     type Error = Error;
 
-    /// Create a CigarString from given bytes.
-    fn try_from(text: &[u8]) -> Result<Self> {
-        Self::try_from(str::from_utf8(text).map_err(|_| Error::BamParseCigar {
-            msg: "unable to parse as UTF8".to_owned(),
-        })?)
-    }
-}
-
-impl TryFrom<&[u8]> for CigarString {
-    type Error = Error;
-
     /// Create a CigarString from given &[u8].
     /// # Example
     /// ```
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -14,9 +14,6 @@ use std::str;
 use std::str::FromStr;
 use std::u32;
 
-use lazy_static::lazy_static;
-use regex::Regex;
-
 use crate::bam::Error;
 use crate::bam::HeaderView;
 use crate::errors::Result;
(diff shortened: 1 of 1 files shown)


dlaehnemann:
That was an early weekend! :D

I kicked of the CI (as you're a first time contributor, :tada: ), and will have a look at the code in a sec.

src/bam/record.rs
```diff
@@ -1126,43 +1123,54 @@ impl TryFrom<&str> for CigarString {
...
+            // check that length is provided
+            if i == j {
                 return Err(Error::BamParseCigar {
-                    msg: "expected cigar operation [0-9]+[MIDNSHP=X]".to_owned(),
+                    msg: "Expected length before cigar operation [0-9]+[MIDNSHP=X]".to_owned(),
                 });
             }
+            // get the length of the operation
+            let n = u32::from_str(&text[i..j]).map_err(|_| Error::BamParseCigar {
+                msg: "Unable to parse to u32".to_owned(),
```
dlaehnemann, line 1143: Maybe make this more specific to ease debugging, something like:
```suggestion
                msg: format!("Unable to parse &str '{}' to u32.", &test[i..j]).to_owned(),
```

src/bam/record.rs
```diff
@@ -1126,43 +1123,54 @@ impl TryFrom<&str> for CigarString {
...
+                    if i == 0 || j + 1 == text_len {
+                        Cigar::HardClip(n)
+                    } else {
+                        return Err(Error::BamParseCigar {
+                            msg: "Hard clipping is only valid at the start or end of a cigar."
+                                .to_owned(),
+                        });
+                    }
+                }
+                "S" => Cigar::SoftClip(n),
```
dlaehnemann, line 1162: Maybe we can also include the specs check for SoftClip(n), this should do the job (although including a test for it is probably a good idea):
```suggestion
                "S" => {
                    if i == 0
                        || j+1 == text_len
                        || &text[i-1..i] == "H"
                        || &text[j+1..].chars().all(|c| c.is_ascii_digit() || c == "H") {
                        Cigar::SoftClip(n)
                    } else {
                        return Err(Error::BamParseCigar {
                        msg: "Soft clips ('S') can only have hard clips ('H') between them and the end of the CIGAR string."
                            .to_owned(),
                        });
                    }
                },
```

dlaehnemann (changes requested):
Really nice and clean code. Thanks a lot!

I have some minor suggestions. And I would opt for removing the old code, even from tests. It's a great idea for testing during the new implementation, but I would rather have those test cases just testing the new code, with hard-coded expected vectors for the `assert_eq!()`. And while writing tests, I would also include one for a hard clip followed by a soft clip, as I did not test my code suggestions... :grimacing:

dlaehnemann (approved):
Perfect, thanks a bunch!

Are you still up for also including a doctest while you're at it? Otherwise we can merge this as is and add one later. Or I'll add it in before merging this.

mrvollger:
Yeah I was able to get to it a little faster than I expected and thanks for your additions. 

The `text[...].char().all(...some test...)` idiom looks super useful. I will have to remember that. 

I added a doc test and it runs on my end.

dlaehnemann:
Found that idiom via [this Stackoverflow answer](https://stackoverflow.com/a/51391684). Generally, there are lots of great Rust questions and answers over there, and the Rust docs then usually provide all the infos I need to know how I can go from there...

I just merged the latest changes from the master branch, and once tests pass we're good to merge.

Thanks again for all the time you took to report this and then to provide such a comprehensive improvement. Hope to see you around for more! :p

tedil:
I only have one more little thing to note here: `text` and `bytes` are used interchangeably here. Since rust strings are always valid utf-8, some characters may be made of more than one byte (as is the case with ascii), e.g. `"fooጷ".as_bytes()` will not have 4 bytes but 6. So if somebody decides to call this with an (obviously invalid) non-ascii cigar string, there will be index errors.

It's probably sufficient to check `text.chars().count() == text.bytes().count()` first.

src/bam/record.rs
```diff
@@ -1124,45 +1121,87 @@ impl TryFrom<&[u8]> for CigarString {
...
+    ///     HardClip(2),
+    /// ]);
+    /// assert_eq!(cigar, expected_cigar);
+    /// ```
     fn try_from(text: &str) -> Result<Self> {
-        lazy_static! {
-            // regex for a cigar string operation
-            static ref OP_RE: Regex = Regex::new("^(?P<n>[0-9]+)(?P<op>[MIDNSHP=X])").unwrap();
-        }
+        let bytes = text.as_bytes();
```
dlaehnemann, line 1145: Following @tedil's [suggestion](https://github.com/rust-bio/rust-htslib/pull/301#issuecomment-829048367):
```suggestion
        let bytes = text.as_bytes();
        if ( text.chars().count() == bytes.len() ) {
            return Err(Error::BamParseCigar {
                msg: "CIGAR string contained non-ASCII characters, which are not valid. Valid are [0-9MIDNSHP=X].".to_owned(),
            });
        }
        
```

natir:
I have a suggestion move all new code in `TryFrom<&[u8]> for CigarString` so all it's one bytes and no question.

And `TryFrom<&str> for CigarString` just perform  check suggest by @tedil and use `&[u8]` implementation.

src/bam/record.rs
```diff
@@ -1124,45 +1121,87 @@ impl TryFrom<&[u8]> for CigarString {
...
+                        msg: format!("Expected cigar operation [MIDNSHP=X] but got [{}]", op),
+                    })
+                }
+            });
+            i = j + 1;
         }
-
         Ok(CigarString(inner))
     }
 }
```
dlaehnemann, line 1207:
```suggestion
}

impl TryFrom<&str> for CigarString {
    type Error = Error;

    /// Create a CigarString from given &str.
    /// # Example
    /// ```
    /// use rust_htslib::bam::record::*;
    /// use rust_htslib::bam::record::CigarString;
    /// use rust_htslib::bam::record::Cigar::*;
    /// use std::convert::TryFrom;
    ///
    /// let cigar_str = "2H10M5X3=2H";
    /// let cigar = CigarString::try_from(cigar_str)
    ///     .expect("Unable to parse cigar string.");
    /// let expected_cigar = CigarString(vec![
    ///     HardClip(2),
    ///     Match(10),
    ///     Diff(5),
    ///     Equal(3),
    ///     HardClip(2),
    /// ]);
    /// assert_eq!(cigar, expected_cigar);
    /// ```
    fn try_from(text: &str) -> Result<Self> {
        let bytes = text.as_bytes();
        if ( text.chars().count() == bytes.len() ) {
            return Err(Error::BamParseCigar {
                msg: "CIGAR string contained non-ASCII characters, which are not valid. Valid are [0-9MIDNSHP=X].".to_owned(),
            });
        }
        CigarString::try_from(bytes)
    }
}

File: src/bam/record.rs (hunk 3 above)
Comment (dlaehnemann) [inline src/bam/record.rs:1145]:
```suggestion
    /// Create a CigarString from given &[u8].
    /// # Example
    /// ```
    /// use rust_htslib::bam::record::*;
    /// use rust_htslib::bam::record::CigarString;
    /// use rust_htslib::bam::record::Cigar::*;
    /// use std::convert::TryFrom;
    ///
    /// let cigar_str = "2H10M5X3=2H".as_bytes();
    /// let cigar = CigarString::try_from(cigar_str)
    ///     .expect("Unable to parse cigar string.");
    /// let expected_cigar = CigarString(vec![
    ///     HardClip(2),
    ///     Match(10),
    ///     Diff(5),
    ///     Equal(3),
    ///     HardClip(2),
    /// ]);
    /// assert_eq!(cigar, expected_cigar);
    /// ```
    fn try_from(text: &[u8]) -> Result<Self> {
```

src/bam/record.rs
```diff
@@ -1779,4 +1818,60 @@ mod alignment_cigar_tests {
             );
         }
     }
+
+    #[test]
+    pub fn test_cigar_parsing_vs_regex() {
+        // parsing test cases
```
dlaehnemann, line 1824: Adding a test for a non-ASCII character in a CIGAR string:
```suggestion
    #[test]
    pub fn test_cigar_parsing_non_ascii_error() {
        let cigar_str = "43ጷ";
        expected_error = Err(Error::BamParseCigar {
                msg: "CIGAR string contained non-ASCII characters, which are not valid. Valid are [0-9MIDNSHP=X].".to_owned(),
            });

        result = CigarString::try_from(cigar_str).expect_err("This should return a BamParseCigar error, but somehow didn't.";
        assert_eq!(expected_error, result);
    }

    #[test]
    pub fn test_cigar_parsing() {
        // parsing test cases
```

dlaehnemann (commented):
I made some suggestions based on the comments from @tedil and @natir. Thanks a bunch for catching this and feel free to comment on the latest changes.

src/bam/record.rs
```diff
@@ -1124,45 +1121,87 @@ impl TryFrom<&[u8]> for CigarString {
...
+    /// Create a CigarString from given &str.
+    /// # Example
+    /// ```
+    /// use rust_htslib::bam::record::*;
+    /// use rust_htslib::bam::record::CigarString;
+    /// use rust_htslib::bam::record::Cigar::*;
+    /// use std::convert::TryFrom;
+    ///
+    /// let cigar_str = "2H10M5X3=2H";
+    /// let cigar = CigarString::try_from("2H10M5X3=2H")
```
dlaehnemann, line 1133:
```suggestion
    /// Create a CigarString from given &[u8].
    /// # Example
    /// ```
    /// use rust_htslib::bam::record::*;
    /// use rust_htslib::bam::record::CigarString;
    /// use rust_htslib::bam::record::Cigar::*;
    /// use std::convert::TryFrom;
    ///
    /// let cigar_str = "2H10M5X3=2H".as_bytes();
    /// let cigar = CigarString::try_from(cigar_str)
```

src/bam/record.rs
```diff
@@ -1124,45 +1121,87 @@ impl TryFrom<&[u8]> for CigarString {
...
+    /// let expected_cigar = CigarString(vec![
+    ///     HardClip(2),
+    ///     Match(10),
+    ///     Diff(5),
+    ///     Equal(3),
+    ///     HardClip(2),
+    /// ]);
+    /// assert_eq!(cigar, expected_cigar);
+    /// ```
     fn try_from(text: &str) -> Result<Self> {
```
dlaehnemann, line 1144:
```suggestion
    fn try_from(text: &[u8]) -> Result<Self> {
```

src/bam/record.rs
dlaehnemann, line 1145:
```suggestion
```

src/bam/record.rs
```diff
@@ -1124,49 +1121,124 @@ impl TryFrom<&[u8]> for CigarString {
...
+    ///     Match(10),
+    ///     Diff(5),
+    ///     Equal(3),
+    ///     HardClip(2),
+    /// ]);
+    /// assert_eq!(cigar, expected_cigar);
+    /// ```
+    fn try_from(text: &str) -> Result<Self> {
+        let bytes = text.as_bytes();
+        if ( text.chars().count() == bytes.len() ) {
```
tedil, line 1233:
```suggestion
        if text.chars().count() != bytes.len() {
```

dlaehnemann:
OK, I wanted to "quickly fix this online", following the suggestions of @tedil and @natir, but things got out of hand. Thanks for picking up the pieces, @tedil! :D

dlaehnemann:
Also, thanks again to @mrvollger for doing the bulk of the work!

mrvollger:
Thanks guys, and I hope to contribute more in the future!

Best,
Mitchell

Issue #134, opened 2019-05-20
Header from BAM file is not parser


slazicoicr:
I downloaded a test BAM file at http://hgdownload.cse.ucsc.edu/goldenPath/hg19/encodeDCC/wgEncodeUwRepliSeq/wgEncodeUwRepliSeqK562G1AlnRep1.bam. Running `samtools view -H` to extract the header produces

```
@SQ	SN:chr1	LN:249250621
@SQ	SN:chr2	LN:243199373
@SQ	SN:chr3	LN:198022430
@SQ	SN:chr4	LN:191154276
@SQ	SN:chr5	LN:180915260
@SQ	SN:chr6	LN:171115067
@SQ	SN:chr7	LN:159138663
@SQ	SN:chr8	LN:146364022
@SQ	SN:chr9	LN:141213431
@SQ	SN:chr10	LN:135534747
@SQ	SN:chr11	LN:135006516
@SQ	SN:chr12	LN:133851895
@SQ	SN:chr13	LN:115169878
@SQ	SN:chr14	LN:107349540
@SQ	SN:chr15	LN:102531392
@SQ	SN:chr16	LN:90354753
@SQ	SN:chr17	LN:81195210
@SQ	SN:chr18	LN:78077248
@SQ	SN:chr19	LN:59128983
@SQ	SN:chr20	LN:63025520
@SQ	SN:chr21	LN:48129895
@SQ	SN:chr22	LN:51304566
@SQ	SN:chrM	LN:16571
@SQ	SN:chrX	LN:155270560
```

However, when I try to extract it with 
```
let bam = Reader::from_path(&"test.bam").unwrap();
let header = Header::from_template(bam.header());
println!("{:?}", header.to_hashmap());
```
an empty HashMap is returned. Further, if I try to save a new BAM file with the header `Writer::from_path(&"out.bam", &header).unwrap();` the resulting BAM file has no header information.

Am I missing something about parsing BAM headers?

slazicoicr:
I figured out that it was the missing `@HD` line that was causing the issue. Adding it causes the code above to parse the header.

I think this is a bug. The specifications at https://samtools.github.io/hts-specs/SAMv1.pdf suggest that the `@HD` be present (Page 9), but it is not mandatory (Page 2). Both `samtools` and [pysam](https://pysam.readthedocs.io/en/latest/api.html) correctly parse the header above.

juliangehring:
[...]

PR #304, opened 2021-05-07
Sync BAM Header Representations

### LINKED ISSUE #156: Certain BAM files cause silent failures or segfaults
# Background

According to the [format specification](https://github.com/samtools/hts-specs/blob/master/SAMv1.pdf), a BAM file can store reference sequence information in *two places* in its headers:

* In a "raw" binary encoding as a list of length `n_ref` comprised of tuples (`l_name`, `name`, `l_ref`)
* In `@SQ` fields inside the plain text SAM header in the `text` field

This is clearly bad. The refseqs stored in these two places should not disagree, but what if they do? The spec doesn't cover that, and it is left up to individual tools and libraries.

* The samtools suite *appears to* merge information from the two sources. (Which takes precedence? I don't know.)
* The Haskell library [Biohazard](https://bitbucket.org/ustenzel/biohazard/) trusts only the binary-encoded refseq headers and completely ignores `@SQ` headers in the `text` field.
* rust-htslib only trusts `@SQ` headers in the `text` field and completely ignores the binary-encoded refseqs, e.g. in the constructor `bam_htslib::bam::Writer::new()`.

# When rust-htslib fails

Consider a trival program that reads in a BAM file, iterates over its records, and writes it to an output BAM file:

```rust
use rust_htslib::bam;
use rust_htslib::bam::Read;

fn main() {
    let mut infile = bam::Reader::from_path("input.bam").unwrap();
    let header = bam::Header::from_template(infile.header());
    let mut outfile = bam::Writer::from_path("output.bam", &header, bam::Format::BAM).unwrap();

    let mut record = bam::Record::new();
    while let Ok(true) = infile.read(&mut record) {
        outfile.write(&record).unwrap();
    };
}
```
If you give this program an input BAM file that was written by a tool based on Biohazard (which has no `@SQ` fields in the SAM header) it will **fail silently**. Its output will be a truncated BAM file containing only the headers, yet it will exit cleanly with a 0 exit code. 😳

If you change the Rust program to use the `bam::Format::SAM` output format instead and give it the same BAM as input, it will segfault.

Note that the input BAM file created by Biohazard is *technically* spec-compliant, and may be processed by samtools. A (ridiculous) workaround is to first roundtrip the input BAM file via SAM using samtools:

```
samtools view -h broken.bam | samtools view -b -o fixed.bam
```

# How should the issue be addressed?

I can see how this could be subject to differing opinions. However, at the very least, *rust-htslib should probably never fail silently(!)*

The rest of the issue isn't so much a bug in rust-htslib as it is the fact that the BAM format isn't clearly specified. My suggestions/questions:

* It would seemingly make sense that rust-htslib to read the refseqs from both possible locations.
* But if they do differ, should it raise an error or simply merge them silently and continue like samtools?
* When rust-htslib writes a BAM file, should it ensure that the refseqs in both header locations are synchronised?

Finally, this issue is probably also the root cause of #134 posted by @slazicoicr.

jch-13:
Fixes #156 and #134.
As discussed in #156 , there are two different representations of reference sequences in BAM headers. `rust-htslib` relies on the `text` representation which has been deprecated by `htslib` in favor of the binary-form representation. The new header API of `htslib` re-builds the textual representation from the binary one on access if necessary. `htslib` deems it necessary if a header-modifying function has been called.

Some BAM file format implementations other than `htslib` (such as [`biohazard`](https://bitbucket.org/ustenzel/biohazard)) do not write textual headers at all, but only trust the binary representation. When a BAM file created by such a library is read by `rust-htslib`, these representations are not synchronized, so we need to call a "header-modifying" function (`sam_hdr_line_index()`) once to be able to get the synchronized textual representation via the appropriate function `sam_hdr_str()` later.


jch-13:
No idea why the CI test failed, it works here...

johanneskoester (approved):
Thanks a lot!

PR #307, opened 2021-05-18
VCF/BCF Writer - auto-recognize format and compression based on file extension

nh13:
This is my first every PR in rust, in any repo, so please forgive my non-idiomatic way of writing Rust.

I was surprised to find that there was no way to create a writer that would auto-recognize the compression and format based on file extension, so here's a first incomplete path.  I am happy to write some tests, clean up the code to be more idiomatic (with help).  But before I do this, I wanted to get the maintainers' thoughts.

### DIFF
--- a/src/bcf/mod.rs
+++ b/src/bcf/mod.rs
@@ -663,6 +663,50 @@ impl Writer {
             Err(Error::NonUnicodePath)
         }
     }
+    
+    /// Create a new writer that writes to the given path.
+    ///
+    /// The compression and format will be determined from the file extension:
+    ///
+    /// * `.vcf`      -> uncompressed VCF
+    /// * `.vcf.gz`.  -> compressed VCF
+    /// * `.vcf.gzip` -> compressed VCF
+    /// * `.vcf.bgzf` -> compressed VCF
+    /// * `.bcf`      -> compressed BCF
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - the path
+    /// * `header` - header definition to use
+    pub fn from_path2<P: AsRef<Path>>(
+        path: P,
+        header: &Header
+    ) -> Result<Self> {
+        if let Some(p) = path.as_ref().to_str() {
+            let fields: Vec<&str> = p.split(".").collect();
+            let length = fields.len();
+            // FIXME: check that we have enough fields
+            let uncompressed: bool = {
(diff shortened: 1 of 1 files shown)


src/bcf/mod.rs
```diff
@@ -663,6 +663,50 @@ impl Writer {
...
+    
+    /// Create a new writer that writes to the given path.
+    ///
+    /// The compression and format will be determined from the file extension:
+    ///
+    /// * `.vcf`      -> uncompressed VCF
+    /// * `.vcf.gz`.  -> compressed VCF
+    /// * `.vcf.gzip` -> compressed VCF
+    /// * `.vcf.bgzf` -> compressed VCF
+    /// * `.bcf`      -> compressed BCF
```
nh13, line 675: Lifted from: https://github.com/samtools/htsjdk/blob/7719274fe370a51a24e6067de21bbe7e18c160a9/src/main/java/htsjdk/samtools/util/FileExtensions.java

src/bcf/mod.rs
```diff
@@ -663,6 +663,50 @@ impl Writer {
...
+    ) -> Result<Self> {
+        if let Some(p) = path.as_ref().to_str() {
+            let fields: Vec<&str> = p.split(".").collect();
+            let length = fields.len();
+            // FIXME: check that we have enough fields
+            let uncompressed: bool = {
+                if (fields[length-1] == "gz" || fields[length-1] == "gzip" || fields[length-1] == "bgzf") { false }
+                else if (fields[length-1] == "bcf") { false }
+                else if (fields[length-1] == "vcf") { true }
+                else { true } // FIXME
```
tedil, line 693: I'd try to use the [`extension` method of Path](https://doc.rust-lang.org/std/path/struct.Path.html#method.extension). It returns `Option<&OsStr>`, which is a bit cumbersome to use since it's no "regular" String/&str, but can be transformed into one, see [here](https://doc.rust-lang.org/std/ffi/struct.OsStr.html).

dlaehnemann:
Good to see you over here! This is definitely a good addition that will make it easier to write VCF/BCF.

As for pointers for more idiomatic Rust, beyond those provided in the code, one great feature of Rust are `doctests`. For all the relevant links, have a look at the rust-bio documentation guidelines:
https://github.com/rust-bio/rust-bio#documentation-guidelines

And finally, if you want some automated recommendations on your code, you could check out [`cargo clippy`](https://github.com/rust-lang/rust-clippy), which is the `Linting` step in the CI.

Also, don't hesitate to ask anything!

PR #308, opened 2021-06-02
Fix compiler warnings

fxwiegand:
Just a quick PR that removes some unused imports and fixes `#[warn(non_fmt_panic)]` similar to https://github.com/rust-bio/rust-bio/pull/430. Also included some clippy lints while i was at it.


brainstorm (approved):
LGTM, thanks for the cleanup!

PR #309, opened 2021-07-05
Make bcf::IndexedReader::fetch functionality inline with htslib regions

mbhall88:
Currently, an end position is required when fetching. This PR changes the fetch method so that `end` is now an `Option<u64>` and setting it to `None` will fetch from `start` until the last record of the contig.

I've also added 
- Some clarifying docs about the fact that the start and end positions are both 0-based inclusive
- More test cases for fetch (including to test the changes)
- Doc example of how to use `bcf::Record::format`


brainstorm (approved):
Makes sense and good accompanying docs and test, thanks for this!

johanneskoester (approved):
Awesome, thank you!

PR #313, opened 2021-07-06
chore: use  PR target event for better handling of PR title checks

johanneskoester:


johanneskoester:
The target event does not seem to trigger anything, and no solution can be found so far. I leave this open for others to jump in here. Does anybody have an idea?

PR #332, opened 2021-07-16
build: update htslib & fix build.rs

pmarks:
@brainstorm this fixes `build.rs` on top of #237.  Fix works on Mac, let's see what happens on other platforms.  This at least keeps us going with the current style of build.rs


brainstorm:
Thanks @pmarks !

Issue #293, opened 2021-01-22
Passing `bam::record::Record` between threads causes a segfault


DonFreed:
Thank you for the very nice library!

I'm working on a tool that passes reads between threads. Unfortunately, the tool is producing non-deterministic segmentation faults and other memory errors. I've traced some of these issues back to rust-htslib, which will crash somewhat randomly when `bam::record::Record`s are passes between threads. I am not too familiar with this library, so I am wondering if this is the expected behavior?

Here is a simplified example that can reproduce the crash:
```rs
use std::error::Error;
use std::str;
use std::sync::mpsc::{self, Receiver};
use std::thread;

use rust_htslib::bam::{Read, Reader};
use rust_htslib::bam::record::Record;

fn sum_mapqs(rx: Receiver<Option<Record>>) -> Result<(), Box<dyn Error>> {
    let mut total_mapq = 0u64;
    loop {
        match rx.recv() {
            Ok(x) => {
                match x {
                    Some(read) => {
                        let mapq = read.mapq();
                        total_mapq = total_mapq.saturating_add(mapq as u64);
                    },
                    None => {  // No more data
                        println!("Total MapQ: {}", total_mapq);
                        return Ok(());
                    },
                }
            },
            Err(e) => {
                eprintln!("Error reciving data: {}", e);
            }
        }
    }
}

fn main() -> Result<(), Box<dyn Error>> {
    let mut bam = match Reader::from_stdin() {
        Ok(bam) => { bam },
        Err(e) => { return Err(Box::new(e)); },
    };

    // Initialize the writer thread
    let (tx, rx) = mpsc::channel();
    let writer = thread::spawn(move || {
        if let Err(e) = sum_mapqs(rx) {
            eprintln!("Error writing output - {}", e);
        }
    });

    for read in bam.records() {
        match read {
            Ok(read) => {
                eprintln!("Parsed read: {}", str::from_utf8(read.qname()).unwrap());
                if let Err(e) = tx.send(Some(read)) {
                    eprintln!("Error sending data to writer thread");
                    return Err(Box::new(e));
                }
            },
            Err(e) => {
                return Err(Box::new(e));
            },
        }
    }

    // Close the spawned thread
    let _ = tx.send(None);
    let _ = writer.join().unwrap();
    Ok(())
}
```

Compiling with `RUSTFLAGS="-g" cargo build` and then running with a SAM passed through stdin produces the following:
```sh
$ cat test.sam | target/debug/rust-htslib-crash
[...]

PR #345, opened 2022-02-02
Workaround Non-Threadsafe Use of `Rc` in `bam::record::Record`

jch-13:
Hey all, this is an attempt to work around #293 (for context please see the discussion there).

I simply removed the reference counted reference to `HeaderView` from BAM `Record`s and impl'ed `genome::AbstractInterval` on a wrapper type around `&Record` and `&HeaderView`. This wrapper type can be constructed like this:
```Rust
let header = bam_reader.header().clone();
for record in bam_reader.records() {
    let mut rec = record.unwrap();
    // `.contig()` is a member of the `genome::AbstractInterval` trait
    let contig = rec.supply_header(&header).contig();
}
```
This is not exactly elegant and I don't know the use cases that made references to headers necessary in the first place, so please let me know if this workaround is viable at all. 

We should definitely try and remove other unsound `unsafe` impls of `Send` and `Sync` for structs containing `Rc` as well.

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -223,7 +223,7 @@ pub trait Read: Sized {
 #[derive(Debug)]
 pub struct Reader {
     htsfile: *mut htslib::htsFile,
-    header: Rc<HeaderView>,
+    header: HeaderView,
     tpool: Option<ThreadPool>,
 }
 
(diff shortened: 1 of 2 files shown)


johanneskoester:
The use case was to make it convenient to conduct header-requiring operations without having the reader at hand (e.g. in a larger codebase, where passing on the reader would make things unnecessarily complicated). @jch-13,  you thing Arc would be an alternative?

PR #349, opened 2022-03-20
feat: Add wrapper of BGZF writer

### LINKED ISSUE #347: Support for BGZF writer?
Would it be possible to add support for writing BGZF files? I'm trying to accelerate a small program that both reads and writes Fastq, and it would be cool to support BGZF on both the input and output sides of things, ideally with multi-threading support.

Currently, it looks like only the Reader from htslib is exposed: https://github.com/rust-bio/rust-htslib/blob/master/src/bgzf/mod.rs

If this isn't feasible, totally understand. If it sounds feasible but you don't have the desire to do it, let me know and I can take a stab at putting a PR together.

jemma-nelson:
This patch exposes BGZF writing capabilities. This is my first time writing unsafe code (outside of tutorials), so some extra attention there would be welcomed. :)

One possibly controversial implementation choice I made:
BGZF supports two different versions of "No Compression." The value `-2` / `wu` results in a "bgzf" writer that simply passes the input through without modifying it - the output file should be the exact same as the input. The value `0` / `w0` results in a more normal/usual "Uncompressed" setting, where the output is conforming bgzf format, but doesn't try to compress it.

I chose to expose both of these, calling the first one "NoCompression" and the latter "Uncompressed." These names aren't clear in how they differ, so I would be happy for them to be renamed - or, if the situation is too confusing, removing NoCompression.

If accepted, this PR should close #347.

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -1034,7 +1034,7 @@ impl Writer {
 ///
 /// * Uncompressed: No compression, zlib level 0
 /// * Fastest: Lowest compression level, zlib level 1
-/// * Maxium: Highest compression level, zlib level 9
+/// * Maximum: Highest compression level, zlib level 9
 /// * Level(i): Custom compression level in the range [0, 9]
 #[derive(Debug, Clone, Copy)]
 pub enum CompressionLevel {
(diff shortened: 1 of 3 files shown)


jemma-nelson:
I was actually unaware of that issue. I'm not sure if it affects us, since I think htslib might be handling that under the hood.

I have a test for the NoCompression level (in which it writes the contents directly, not BGZF encoded) in `fn test_write_plain()`, but it could probably have a more obvious name. :)

PR #354, opened 2022-05-09
fix: Update Cargo.toml

rob-p:
Bump version in Cargo.toml file to match the latest actual release.


johanneskoester:
Thanks! Closing in favor of https://github.com/rust-bio/rust-htslib/pull/356. I had to push the version bump there manually, although the release please action is supposed to do that. For some reason, it does it for hts-sys-releases but not for rust-htslib. This might be a bug in release please.

PR #366, opened 2022-10-12
perf: update htslib and corresponding bindings to 1.16

johanneskoester:


johanneskoester:
@pmarks could you help me here? I must admit that I have no idea anymore about the build process of hts-sys. As far as I remember, you redesigned it some time ago. In case something in the rust code needs to be adapted to API changes, let me know. But currently, the errors rather look like I am doing something wrong with the update of htslib and the bindings itself.

johanneskoester:
Also, while I tried to manually generate the prebuilt bindings for linux, I cannot do that for osx, since I don't have an apple machine here.

pmarks:
@johanneskoester I think this is good to go.

johanneskoester:
Awesome, thanks a lot!

PR #374, opened 2023-02-13
feat: Implement htslib basemod api

jts:
This PR adds basic support for base modifications using htslib's internal API. The modifications are accessed through a new `base_modifications()` function added to `Record`, which returns an iterable struct. The associated functions allocate memory needed to track the internal state, so I'd appreciate a review to make sure my implementation is consistent with other parts of the library.

One design note: the number of modifications at each position is variable, so the `basemod_position_iter` returns a new vector at each step (internally htslib writes into a buffer the caller provides). The basic `basemod_iter` will return a single modification each time, with consecutive `next()` calls potentially returning modifications for the same position of SEQ. 

(02/17 updated this comment after revising the iterator design in `fe6b21f`)

### DIFF
--- a/hts-sys/linux_prebuilt_bindings.rs
+++ b/hts-sys/linux_prebuilt_bindings.rs
@@ -9329,6 +9329,129 @@ extern "C" {
         flag: ::std::os::raw::c_int,
     ) -> ::std::os::raw::c_int;
 }
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct hts_base_mod_state {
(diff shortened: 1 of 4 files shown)


jts:
@johanneskoester could you have a look at this PR?

johanneskoester:
hts-sys was moved into a separate repo: https://github.com/rust-bio/hts-sys
May I ask you to make the binding modifications there?

jts:
I am closing this PR in favor of a new one that is rebased to the current master.

PR #375, opened 2023-02-23
feat: Revised calculation of leading- and trailing-softclips

FelixMoelder:
The `leading_softclips` and `trailing_softclips` functions of `CigarStringView` objects return 0 softclips in case of hardclipping operations between the softclips and the end of the cigar string e.g `50M5S10H`.
While it is true that the example Cigar is not concluded by a softclip and therefore returns 0 softclips I would argue that the functions should always returns the number of softclips on the corresonding end no matter if there are hardclips or not.
As softclips are part of a records sequence they need to be taken into account when determining unaligned read positions independently of hardclips.

Note:
As I implemented a recursive call of `leading/trailing_softclip` functions I had to rebuild a new CigarStringView when I just set the position to 0. I do not see any issue with that but just want to point that out in case one comes up with a less sloppy implementation.

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -1912,7 +1912,7 @@ impl<'a> IntoIterator for &'a CigarString {
     type IntoIter = ::std::slice::Iter<'a, Cigar>;
 
     fn into_iter(self) -> Self::IntoIter {
-        (&(self.0)).iter()
+        self.0.iter()
     }
 }
 
@@ -1927,13 +1927,12 @@ impl fmt::Display for CigarString {
 
 // Get number of leading/trailing softclips if a CigarString taking hardclips into account
 fn calc_softclips<'a>(mut cigar: impl DoubleEndedIterator<Item = &'a Cigar>) -> i64 {
-    let softclips = match (cigar.next(), cigar.next()) {
+    match (cigar.next(), cigar.next()) {
         (Some(Cigar::HardClip(_)), Some(Cigar::SoftClip(s))) | (Some(Cigar::SoftClip(s)), _) => {
             *s as i64
         }
         _ => 0,
-    };
-    softclips
+    }
 }
 
 #[derive(Eq, PartialEq, Clone, Debug)]
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -1960,24 +1960,26 @@ impl CigarStringView {
 
     /// Get number of bases softclipped at the beginning of the alignment.
     pub fn leading_softclips(&self) -> i64 {
-        self.first().map_or(0, |cigar| {
-            if let Cigar::SoftClip(s) = cigar {
-                *s as i64
-            } else {
-                0
-            }
-        })
+        self.split_first()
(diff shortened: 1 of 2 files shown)


johanneskoester (changes requested):
Thanks a lot! Instead of constructing the cigar string again, I would have an internal function that takes a cigar string and returns the leading/trailing softclips. This can then be used for both cases (with hardclip or not)

FelixMoelder:
> Thanks a lot! Instead of constructing the cigar string again, I would have an internal function that takes a cigar string and returns the leading/trailing softclips. This can then be used for both cases (with hardclip or not)

I moved everything to a internal function being called by the both public functions.

src/bam/record.rs
```diff
@@ -1774,6 +1775,15 @@ impl CigarString {
 
         CigarString(cigar)
     }
+
+    // Get number of leading/trailing softclips taking hardclips into account
+    fn get_softclips(self) -> i64 {
```
johanneskoester, line 1780: This should use &self, so that you don't need to take ownership.

johanneskoester: Also, I would name it differently, like `calc_leading_softclips`.

johanneskoester (changes requested):

src/bam/record.rs
```diff
@@ -1970,14 +1970,14 @@ impl CigarStringView {
 
     /// Get number of bases softclipped at the beginning of the alignment.
     pub fn leading_softclips(&self) -> i64 {
-        self.to_owned().take().get_softclips()
+        self.to_owned().take().calc_leading_softclips()
```
johanneskoester, line 1973: I don't understand why this would still need to_owned and take.

src/bam/record.rs
```diff
@@ -1970,14 +1970,14 @@ impl CigarStringView {
...
     }
 
     /// Get number of bases softclipped at the end of the alignment.
     pub fn trailing_softclips(&self) -> i64 {
         let mut cigar = self.to_owned().take();
         cigar.reverse();
-        cigar.get_softclips()
+        cigar.calc_leading_softclips()
```
johanneskoester, line 1980: I just noticed, when you turn the calc_leading_softlclips function into one taking a [DoubleEndedIterator](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html), you can just pass either the normal iterator or the reversed one without having to copy anything.

johanneskoester (changes requested):

src/bam/record.rs
FelixMoelder: `&self` is a reference of a `CigarStringView but we need a `CigarString`-object/reference when calling `calc_leading_softclips`. But there is no function returning a reference of a `CigarString` which is why I need to own the reference first.

src/bam/record.rs
FelixMoelder: Good point. This should be fine now.

FelixMoelder:
@johanneskoester For some reasons the CI is not running any test and just passes by  checking the title.

johanneskoester (approved):

PR #377, opened 2023-02-23
feat: Add ability to fetch number of sequences and I-th sequence from FAI index

jamorrison:
`faidx.c` has two functions for retrieving the number of sequences in the FAI index (`faidx_nseq`) and the name of the I-th sequence (`faidx_iseq`). Previously these two functions weren't accessible through `rust-htslib`. I've added a function to access each of these into the `faidx` module.

This is my first PR for `rust-htslib`, so let me know if there are any things I can improve upon in my code!

### DIFF
--- a/src/faidx/mod.rs
+++ b/src/faidx/mod.rs
@@ -111,7 +111,7 @@ impl Reader {
     /// # Arguments
     ///
     /// * `i` - index to query
-    pub fn fetch_i_seq(&self, i: i32) -> Result<String> {
+    pub fn seq_name(&self, i: i32) -> Result<String> {
         let cname = unsafe {
             let ptr = htslib::faidx_iseq(self.inner, i);
             ffi::CStr::from_ptr(ptr)
--- a/src/faidx/mod.rs
+++ b/src/faidx/mod.rs
@@ -101,7 +101,7 @@ impl Reader {
     }
 
     /// Fetches the number of sequences in the fai index
-    pub fn fetch_n_seqs(&self) -> u64 {
+    pub fn n_seqs(&self) -> u64 {
         let n = unsafe { htslib::faidx_nseq(self.inner) };
         n as u64
     }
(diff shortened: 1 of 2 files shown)


src/faidx/mod.rs
```diff
@@ -99,6 +99,33 @@ impl Reader {
         let bytes = self.fetch_seq(name, begin, end)?;
         Ok(std::str::from_utf8(bytes).unwrap().to_owned())
     }
+
+    /// Fetches the number of sequences in the fai index
+    pub fn fetch_n_seqs(&self) -> u64 {
```
johanneskoester, line 104:
```suggestion
    pub fn n_seqs(&self) -> u64 {
```

src/faidx/mod.rs
```diff
@@ -99,6 +99,33 @@ impl Reader {
...
+        let n = unsafe { htslib::faidx_nseq(self.inner) };
+        n as u64
+    }
+
+    /// Fetches the i-th sequence name
+    ///
+    /// # Arguments
+    ///
+    /// * `i` - index to query
+    pub fn fetch_i_seq(&self, i: i32) -> Result<String> {
```
johanneskoester, line 114:
```suggestion
    pub fn seq_name(&self, i: i32) -> Result<String> {
```

johanneskoester (changes requested):
Thanks! I would rename the functions a bit, see below.

brentp:
I need this functionality, anything holding up the PR?

brentp:
Thanks @brainstorm . 
@jamorrison , you can take my tests nearly verbatim from #381

jamorrison:
Thanks @brentp for the test code!

@johanneskoester I've added added tests for the two functions in the PR. Let me know if there's further work that needs to be done to wrap up this PR.

johanneskoester (approved):

johanneskoester (approved):

PR #381, opened 2023-03-29
faidx: expose n_seq and seqi

brentp:


brentp:
close in favor of #377

PR #384, opened 2023-04-10
fix: Allow empty tags in BAM header

mrvollger:
Hello, 

I recently ran into an issue where BAM files with empty tags in their header fail with rust-htslib whereas they work fine with samtools and even with `quickcheck` within samtools. For example a barcode tag without a value (`BC:`) within `@RG` would throw an unwrap error when using `to_hashmap`. 

This just needs a one character change in the regex `+` -> `*`, so hopefully wont cause any issues or changes in compatibility. 

Thanks for considering!
Mitchell


johanneskoester:
thanks a lot!

Issue #386, opened 2023-04-26
Accessing the index file from a `bam::IndexedReader`


tedil:
I noticed that there was neither a non-unsafe way of accessing the actual index of an `IndexedReader` nor any "convenience" methods related to the index.

Therefore, I've tried translating [`samtools idxstats`](https://github.com/samtools/samtools/blob/afe7fb30055080f43f114e51594a9c2f9a337e98/bam_index.c#L263-L274) using the htslib FFI interface, but stumbled upon the following issue:

When accessing the index of an `IndexedReader`, the result is different than when loading the index manually through `sam_index_load`, as can be reproduced with the following code and the example BAM file and index in [data.zip](https://github.com/rust-bio/rust-htslib/files/11331614/data.zip).

```rust
use std::ffi;

use rust_htslib::bam;
use rust_htslib::bam::{FetchDefinition, Read};
use rust_htslib::htslib::hts_idx_t;

/// Get length, mapped and unmapped read counts for each contig.
unsafe fn idxstats<R: bam::Read>(
    bam: &mut R,
    idx: *mut hts_idx_t,
) -> Vec<(String, (u64, u64, u64))> {
    bam.header()
        .target_names()
        .iter()
        .enumerate()
        .map(|(tid, tname)| {
            let (mut mapped, mut unmapped) = (0, 0);
            rust_htslib::htslib::hts_idx_get_stat(idx, tid as i32, &mut mapped, &mut unmapped);
            let tname = std::str::from_utf8(tname).unwrap();
            let tlen = bam.header().target_len(tid as u32).unwrap();
            (tname.to_string(), (tlen, mapped, unmapped))
        })
        .collect()
}

fn main() {
    let mut reader = bam::IndexedReader::from_path("data/test.bam").unwrap();
    reader.fetch(FetchDefinition::All).unwrap();

    let stats_from_indexed_reader = unsafe {
        let idx = (*reader.htsfile()).idx;
        idxstats(&mut reader, idx)
    };

    let stats_from_index = unsafe {
        let path = "data/test.bam.bai";
        let c_str = ffi::CString::new(path).unwrap();
        let htsfile = reader.htsfile();
        let idx = rust_htslib::htslib::sam_index_load(htsfile, c_str.as_ptr());
        idxstats(&mut reader, idx)
    };

    assert_eq!(stats_from_indexed_reader, stats_from_index);
}
```

[...]

PR #387, opened 2023-05-02
feat: Index for `bam::IndexedReader`

tedil:
This PR addresses issue #386 by providing a struct `IndexView` analogous to `HeaderView`.

Since I'm not particularly well versed in interfacing with C and the htslib API, I'm a bit unsure if this is "the correct way" to handle this. There's no corresponding implementation for `Clone` yet, and I'm not sure what the point of the `owned` field of `HeaderView` is, since it does not seem to be modified anywhere.

~~This still needs some testcases, a simple one should be the one from the referenced issue.~~


tedil:
This currently only works for BAM + bai, not CRAM + crai, unsure why that is.

tedil:
Ah, the cram index does not contain the necessary information, so the slow idxstats variant has to be used, see [bam_index.c](https://github.com/samtools/samtools/blob/556c60fdff977c0e6cadc4c2581661f187098b4d/bam_index.c#L140-L150) for more information

tedil:
I've added an unpolished translation of the slow_idxstats method from [samtools bam_index.c](https://github.com/samtools/samtools/blob/556c60fdff977c0e6cadc4c2581661f187098b4d/bam_index.c#L140-L199), which should be used in case of CRAM (and SAM) indexed files.

johanneskoester (approved):

johanneskoester:
Nice work! If I remember correctly, the owned flag was for destroying the header view, which is not always what you want. But I don't recall the details anymore and don't have the time to check the code. However, this one looks good to me for now.

PR #388, opened 2023-05-05
feat: HeaderRecord::push_tag: Value may be owned

sjackman:
The type of the value may be a ref or owned. Previously it must be a ref.


johanneskoester (approved):

sjackman:
Thanks for merging, Johannes!

Issue #276, opened 2020-09-30
BAM header to_hashmap() fails on CO records


andrewpatto:
Just in doing some work on printing example BAM headers - and thought we'd use to_hashmap() to get us the header data nicely split up. But it appears to be panicking on the unwrap() at line 88

`let cap = TAG_RE.captures(part).unwrap();`

which would make sense as the `TAG_RE` expects to see a colon, but the comment records do not necessarily fit that pattern

For example, the public cram

`"s3://1000genomes/data/NA18616/alignment/NA18616.alt_bwamem_GRCh38DH.20150718.CHB.low_coverage.cram"`

has some header records

"@CO\tFASTQ=ERR009378_1.fastq.gz",

without investigating any further, my guess is that those are the records that are failing. I'm not enough of a bioinformatics expect to know if it is the headers themselves at fault, so I'll just leave it here as an issue.

andrewpatto:
@brainstorm follow up to the lambda example

PR #396, opened 2023-06-14
feat!: bam header to_hashmap method now returns a result in order to deal with header parsing issues

ArtRand:
I've recently encountered an issue where the [to_hashmap](https://github.com/rust-bio/rust-htslib/blob/e89538d5a9971c6508ac38d92ac468f3d70241aa/src/bam/header.rs#L70) method will cause a panic ([issue](https://github.com/nanoporetech/modkit/issues/29)). #276 reports a similar problem. I've refactored the function to return a `Result` allowing the caller to handle the failure. I understand this changes the API and it may be considered unconventional to have a `to_*` method return a result, so I'm open to discussion about other ways to aviod the panic.


tedil:
An alternative would be to implement [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) (i.e. `impl TryFrom<Header> for HashMap<String, String>`), such that
```rust
let header_map: HashMap<_, _> = header.try_into()?;
```
API wise, that would be more difficult to discover, but it does leverage existing traits and clearly signals that something during conversion might fail.

(Also, instead of the `if let`, one *could* use a combination of `.map` and `.map_err` instead, but that boils down to personal preference I guess ;) )

veldsla:
This possible panic was already fixed in #384 (released in version 0.43.0) header tags may now be empty.

ArtRand:
@tedil 

I'm happy to implement it as a `TryFrom` if you think that's best, I tend to agree that discovery would be difficult. 

@veldsla 

I appreciate that this particular error would not cause a panic with the new regex pattern. However, we've all experienced out-of-spec data. At the very least I'd expect the method's documentation to mention that it can cause a panic when the header isn't to the spec and (IMO) it's much more ideal to communicate a potential failure with a `Result` type. 

I've refactored this function call out of my project, but would be more than happy to contribute a more reliable solution.

johanneskoester (approved):

johanneskoester:
My feeling is that to_hashmap is more ergonomic here. Let's keep it for now. Makes sense to return a result here.

PR #397, opened 2023-06-20
fix: include doctests (and code examples) in test coverage calculations

dlaehnemann:


dlaehnemann:
I'm not sure why coverage seems to decrease in some places, but it does seem to increase (marginally) overall. And looking at the tarpaulin run in the CI, it is definitely running both the regular `mod tests` and the `doctests`. So I think this should improve on the previous coverage reporting.

johanneskoester (approved):

PR #398, opened 2023-06-20
fix: use correct return value in bcf_get_format and bcf_get_info_values

Marlin-Na:
According to the [notes in htslib vcf header file](https://github.com/samtools/htslib/blob/7f69840c2fbf73dc7601c17a3ade4db676858cf4/htslib/vcf.h#L1118-L1120):

> Use the returned number of written values for accessing valid entries of dst, as ndst is only a
> watermark that can be higher than the returned value, i.e. the end of dst can contain carry-over
> values from previous calls to bcf_get_format_*() on lines with more values per sample.

I had a case that when accessing a format field of string type, even there was only one sample, I got two strings returned (albeit the second one was empty). With this change, the issue was fixed.

### DIFF
--- a/src/bcf/mod.rs
+++ b/src/bcf/mod.rs
@@ -1027,12 +1027,7 @@ mod tests {
                 .string()
                 .expect("Error reading string.");
             assert_eq!(fs1_str_vec.len(), 2);
-            println!(
-                "{}",
-                String::from_utf8_lossy(
-                    fs1_str_vec[0]
-                )
-            );
+            println!("{}", String::from_utf8_lossy(fs1_str_vec[0]));
             assert_eq!(
                 record
                     .format(b"FS1")
--- a/src/bcf/mod.rs
+++ b/src/bcf/mod.rs
@@ -1022,13 +1022,15 @@ mod tests {
                     .expect("Missing tag")[0],
                 format!("string{}", i + 1).as_bytes()
             );
+            let fs1_str_vec = record
+                .format(b"FS1")
+                .string()
+                .expect("Error reading string.");
+            assert_eq!(fs1_str_vec.len(), 1);
             println!(
                 "{}",
                 String::from_utf8_lossy(
-                    record
-                        .format(b"FS1")
-                        .string()
-                        .expect("Error reading string.")[0]
+                    fs1_str_vec[0]
                 )
             );
             assert_eq!(
--- a/src/bcf/mod.rs
+++ b/src/bcf/mod.rs
@@ -1026,7 +1026,7 @@ mod tests {
                 .format(b"FS1")
                 .string()
                 .expect("Error reading string.");
-            assert_eq!(fs1_str_vec.len(), 1);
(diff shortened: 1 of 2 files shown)


johanneskoester (approved):

johanneskoester (changes requested):
Can you please add one or two test cases which show that ret is the right thing to use and also works in case of multiple strings?

Marlin-Na:
@johanneskoester I added one test for this. It will fail on master but pass on this PR.

dlaehnemann:
Another thing I noticed: It seems like we then simply have `ret` returned once as a `usize` and once as an `i32`. And indeed, most of the calls to these internal functions seem to use only one version. And in the other cases, one could simply cast it into the other type. So maybe just return `Ok(Some(ret))` as is and then cast it to `usize` whenever that is necessary?

Marlin-Na:
Looking at the code again. I think maybe an alternative fix is to change
https://github.com/rust-bio/rust-htslib/blob/261e3a3b027c84490a9a2faa29839a91fc5f78ba/src/bcf/record.rs#L1473
in `Format` reader to use `ret` instead of `n`.

The `Info` reader actually got it correct:
https://github.com/rust-bio/rust-htslib/blob/261e3a3b027c84490a9a2faa29839a91fc5f78ba/src/bcf/record.rs#L1329

Marlin-Na:
But `float()` and `integer()` in `Format` read also got it wrong. Yeah I think we should just use `ret` instead of `n`. I will implement the fix @dlaehnemann  suggested.
https://github.com/rust-bio/rust-htslib/blob/261e3a3b027c84490a9a2faa29839a91fc5f78ba/src/bcf/record.rs#L1455

src/bcf/mod.rs
```diff
@@ -1022,13 +1022,15 @@ mod tests {
                     .expect("Missing tag")[0],
                 format!("string{}", i + 1).as_bytes()
             );
+            let fs1_str_vec = record
+                .format(b"FS1")
+                .string()
+                .expect("Error reading string.");
+            assert_eq!(fs1_str_vec.len(), 2);
```
Marlin-Na, line 1029: This checks the string vector length is consistent with sample count. However I couldn't easily reproduce the the case I had where `n != ret`, as it happened in the middle of reading a large vcf file and the criteria is unknown.

dlaehnemann: From what I understand from the docs you cited, `n` is really a watermark. So you probably had a record with two string entries before the record with one string entry where the problem occurred.

So maybe you can change your test to actually read in a `.vcf` file with two records. The first one with to string entries, the second record with only one string entry. I think that this should then trigger that behaviour.

Marlin-Na: I think I know the issue. The test needs to reuse the shared buffer to trigger the behavior. I have updated the tests. Now they will fail on master and pass with this PR.

Marlin-Na:
I think this PR is ready for merging.

johanneskoester (approved):

PR #408, opened 2023-09-28
feat: Add rust_htslib::bcf::index::build

PB-DB:
Builds a bcf/vcf index using the htslib bindings. It was useful to us, so I'm happy to share.

It's modeled after the equivalent in `rust_htslib::bam::index`. I've used it myself to verify behavior, but I haven't built any doc or unit tests.

Hope it's helpful!

Best,

Daniel

### DIFF
--- a/src/bcf/index.rs
+++ b/src/bcf/index.rs
@@ -43,6 +43,26 @@ impl std::fmt::Display for BcfBuildError {
 
 /// Build a bcf or vcf.gz index.
 /// Builds tbi or csi depending on index_type.
+///
+///```
+/// // Index a sorted bcf file with csi.
+/// let bcf_path = concat!(env!("CARGO_MANIFEST_DIR"), "/test/test_multi.bcf");
+/// rust_htslib::bcf::index::build(&bcf_path, Some(&"built_test.csi"), /*n_threads=*/ 4u32, rust_htslib::bcf::index::Type::Csi(14)).expect("Failed to build csi index for bcf file.");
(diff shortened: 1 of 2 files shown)


johanneskoester (approved):
Thanks!

johanneskoester (changes requested):
Can you please add a test case?

PB-DB:
Sure!

I added some basic tests for successful construction from bcf + vcf.gz, as well as assert errors on building the wrong index format.

It would be convenient if it automatically chose 14 for a min shift for CSI files, or if we autodetected the index type to build, but I'm not sure it would make sense for the library. I'm happy to if you think it's worth it.

Thanks for building such a great resource!

johanneskoester:
> Sure!
> 
> I added some basic tests for successful construction from bcf + vcf.gz, as well as assert errors on building the wrong index format.
> 
> It would be convenient if it automatically chose 14 for a min shift for CSI files, or if we autodetected the index type to build, but I'm not sure it would make sense for the library. I'm happy to if you think it's worth it.
> 
> Thanks for building such a great resource!

Sure, feel free to extend the functionality in that direction in a follow-up PR! And thanks for liking rust-htslib!

johanneskoester:
May I ask you to update to merge in the latest upstream master branch? It contains fixes for the failing tests.

PB-DB:
Done - and thanks for figuring that out. Hope this fixes the tests! 🤞

johanneskoester (approved):

Issue #412, opened 2023-11-28
 error[E0308]: mismatched types - expected raw pointer `*mut u8` during cross compilation


Adoni5:
Hey guys - I'm trying cross compile into `aarch64-unknown-linux`. The problem is the bindings seem to expect `*mut u8`, and are receiving an `i8` reference.
```console
 error[E0308]: mismatched types
    --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rust-htslib-0.44.1/src/bam/record.rs:2324:17
     |
2319 |             let ret = hts_sys::bam_mods_query_type(
     |                       ---------------------------- arguments to this function are incorrect
...
2324 |                 &mut canonical,
     |                 ^^^^^^^^^^^^^^ expected `*mut u8`, found `&mut i8`
     |
     = note:    expected raw pointer `*mut u8`
             found mutable reference `&mut i8`
note: function defined here
    --> /home/runner/work/cnv_from_bam/cnv_from_bam/target/aarch64-unknown-linux-gnu/release/build/hts-sys-f85a33d3986fc8a3/out/bindings.rs:9439:12
     |
9439 |     pub fn bam_mods_query_type(
     |            ^^^^^^^^^^^^^^^^^^^
```

Specifically the issue seems to be here in the modified base metadata.

https://github.com/rust-bio/rust-htslib/blob/3008a131f241b423d041c756fc96410f6412e3d8/src/bam/record.rs#L2317-L2325

Am I doing something obviously wrong? I'm cross compiling using an altered version of `docker/cross`

https://github.com/rust-cross/manylinux-cross/pkgs/container/manylinux_2_28-cross. I'm creating a `.so` to use with a `PyO3` package.

Many Thanks
Rory

PR #415, opened 2024-01-02
fix: Loosen acceptable types to support current linux build on aarch64

rdeaton-freenome:
Fixes https://github.com/rust-bio/rust-htslib/issues/412

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -4,6 +4,7 @@
 // except according to those terms.
 
 use std::convert::TryFrom;
+use std::convert::TryInto;
 use std::ffi;
 use std::fmt;
 use std::marker::PhantomData;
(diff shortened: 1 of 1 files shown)


ghuls:
Same for `ppc64le`.

johanneskoester (commented):
aarch64 does not support i8?

ghuls:
@johanneskoester I assume it is because of this: https://abstractexpr.com/2023/04/30/the-anomaly-of-the-char-type-in-c/

rdeaton-freenome:
That article is spot on, it is fundamentally because the signedness of char is different across platforms. There are definitely better ways to fix this long term, but luckily the range of values here is so small that taking half of the i8 range is sufficient. Thanks!

Issue #407, opened 2023-08-17
Panic on omitted trailing FORMAT fields


fennerm:
I'm running into a panic when attempting to parse a VCF with rust_htslib::bcf. I can't share the real VCF but here's a minimal example:
```
##fileformat=VCFv4.3
##contig=<ID=chr1,length=10000>
##INFO=<ID=FOO,Number=1,Type=Integer,Description="Some field">
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
##FORMAT=<ID=ABC,Number=1,Type=String,Description="Some string field">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	PrecisionFDA_HG002
chr1	1234	.	tttatttatttatttatttatttattta	ttta	.	.	FOO=1	GT:ABC	.
```

The problem is that the sample column only has one "." but there are two fields defined in the FORMAT column. Per the VCF spec I _think_ this is valid:
> Trailing fields can be dropped, with the exception of the GT field, which should always be present if specified in the FORMAT
field.

Panic message:
```
thread panicked at 'chunk size must be non-zero', /Users/fennerm/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rust-htslib-0.44.1/src/bcf/record.rs:1490:18
```
Relevant line in code:
https://github.com/rust-bio/rust-htslib/blob/master/src/bcf/record.rs#L1490

fennerm:
Will try to debug a bit and submit a PR

fennerm:
Took a closer look but I think this one is a bit beyond me. Jotting down my notes:
* Error only occurs if the FORMAT field has type=String - Float/Integer work fine.
* `bcf_get_format_values` and `self.values_per_sample()` are both 0 if the field has type=String 
    * this causes the panic in `.chunks()`)
* `bcf_get_format_values` and `self.values_per_sample()` are both 1 if the field has type=Integer
* Unsure if that inconsistency is expected

PR #417, opened 2024-02-01
Handle trailing omitted FORMAT records

fennerm:
Fix for https://github.com/rust-bio/rust-htslib/issues/407

### DIFF
--- a/src/bcf/record.rs
+++ b/src/bcf/record.rs
@@ -1329,6 +1329,9 @@ impl<'a, 'b, B: BorrowMut<Buffer> + Borrow<Buffer> + 'b> Info<'a, B> {
     pub fn string(mut self) -> Result<Option<BufferBacked<'b, Vec<&'b [u8]>, B>>> {
         self.data(htslib::BCF_HT_STR).map(|data| {
             data.map(|ret| {
+                if ret == 0 {
+                    return BufferBacked::new(Vec::new(), self.buffer);
+                }
                 BufferBacked::new(
                     unsafe {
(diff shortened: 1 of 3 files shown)


src/bcf/record.rs
```diff
@@ -1329,6 +1329,9 @@ impl<'a, 'b, B: BorrowMut<Buffer> + Borrow<Buffer> + 'b> Info<'a, B> {
     pub fn string(mut self) -> Result<Option<BufferBacked<'b, Vec<&'b [u8]>, B>>> {
         self.data(htslib::BCF_HT_STR).map(|data| {
             data.map(|ret| {
+                if ret == 0 {
+                    return BufferBacked::new(Vec::new(), self.buffer);
```
fennerm, line 1333: I'm not sure if this is the behavior we want here. Could consider returning `"."` but then I guess we'd need to modify the buffer

src/bcf/record.rs
```diff
@@ -1483,6 +1483,9 @@ impl<'a, 'b, B: BorrowMut<Buffer> + Borrow<Buffer> + 'b> Format<'a, B> {
     /// memory.
     pub fn string(mut self) -> Result<BufferBacked<'b, Vec<&'b [u8]>, B>> {
         self.data(htslib::BCF_HT_STR).map(|ret| {
+            if ret == 0 {
```
fennerm, line 1486: I'm not sure if this is the behavior we want here. Could consider returning "." but then I guess we'd need to modify the buffer

johanneskoester: empty seems fine

johanneskoester (approved):

PR #431, opened 2024-05-03
fix: build for macOS

FelixMoelder:
For getting the CI running on `aarch64-apple-darwin` runners hts-sys needs to be updated due to an issue caused by bindgen. This has been fixed in the latest `hts-sys` release (see https://github.com/rust-bio/hts-sys/pull/10).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ bio-types = ">=0.9"
 byteorder = "1.3"
 custom_derive = "0.1"
 derive-new = "0.5"
-hts-sys = {version = "2.1.1", default-features = false}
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -21,7 +21,7 @@ bio-types = ">=0.9"
 byteorder = "1.3"
 custom_derive = "0.1"
 derive-new = "0.5"
-hts-sys = {version = "2.1.1", default-features = false}
+hts-sys = {version = "2.1.4", default-features = false, features = ["bindgen"]}
```
sjackman, line 24: Was this an intentional change that `bindgen` changed from an optional dependency to a required dependency of `rust-htslib`?

PR #435, opened 2024-07-08
ci: Add Dependabot

sjackman:
Enable [Dependabot](https://docs.github.com/en/code-security/dependabot) to open PRs to update dependencies.

Related PR in `rust-bio/rust-bio`:

- <https://github.com/rust-bio/rust-bio/pull/536>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      # Prefix all commit messages with "deps: ", which should be
+      # accepted as a conventional commit and trigger release-please
+      prefix: "deps"


sjackman:
@nh13 Hi, Nils. Could you please review and merge this PR to enable Dependabot to update dependencies in `rust-htslib`? Dependabot is being using in both `bio` and `bio-types`.

nh13:
@sjackman I am not a maintainer unfortunately. CC: @FelixMoelder

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      # Prefix all commit messages with "deps: ", which should be
+      # accepted as a conventional commit and trigger release-please
+      prefix: "deps"
```
fxwiegand, line 10: I don't think this is necessary. Check out our [Dependabot configuration for datavzrd](https://github.com/datavzrd/datavzrd/blob/main/.github/dependabot.yml) (another software of ours). This is not specified and the PRs are titles with _build(deps):_. I don't think _deps:_ is part of the angular convention that conventional commits and release-please rely on.

```suggestion
```

fxwiegand (changes requested):
Hi and thanks for adding Dependabot! Felix and Johannes are both on vacation at the moment. Just a small change before we can get this merged ☺️

.github/dependabot.yml
sjackman: `deps:` is the convention used by both `rust-bio` and `rust-bio-types`:
https://github.com/rust-bio/rust-bio/blob/f726f3511f7286b687687b811441599b98f1dc82/.github/dependabot.yml#L7-L10
https://github.com/rust-bio/rust-bio-types/blob/e36648bd3a0f8c5cd711d060bd647f369a2c954e/.github/dependabot.yml#L7-L10

Can you please confirm that `rust-bio/rust-htslib` wishes to use a different convention `build(deps):` than the convention `deps:` used by both `rust-bio` and `rust-bio-types`?

johanneskoester: We can keep it like this.

PR #440, opened 2024-09-30
buadded SamFlag module

rLannes:
Description:

This pull request introduces a new module, `bam::sam_flag`, to facilitate easier handling of SAM flags. The module provides a more intuitive and type-safe way to work with SAM flags, improving code readability and reducing potential errors.

Key features:

1. `SamFlag` struct: A zero-cost abstraction that contains constants for all standard SAM flags.
2. `check_flag()` function: A utility function to easily check for the presence or absence of specific flags.

Benefits:
- Improved code readability when working with SAM flags.
- Reduced chance of errors from manual flag manipulation.
- Type-safe constants for all standard SAM flags.
- Easy-to-use function for checking flag combinations.

Implementation choice:
The main design choice was to use a struct with associated constants rather than an enum to represent the flag values. The benefit is no runtime cost, but it comes at the cost of not being as nice to work with as an enum.

Example usage:

```rust
let read_flag = 65; // PAIRED (1) + FIRST_IN_PAIR (64)
assert!(check_flag(read_flag, SamFlag::PAIRED + SamFlag::FIRST_IN_PAIR, SamFlag::READ_UNMAPPED));
```

### DIFF
--- a/src/bam/sam_flag.rs
+++ b/src/bam/sam_flag.rs
@@ -1,0 +1,87 @@
+
+
+//! A module that provides an easier way to work with SAM flags.
+//! It achieves this by providing a struct (`SamFlag`) with associated constants representing 
+//! flags, and the `check_flag()` function that allows testing if specific flags are set or not set.
+//! 
+//! For example, the following code tests if the read flag has the FIRST_IN_PAIR flag set and the MATE_UNMAPPED flag not set:
+//! ```
+//! use rust_htslib::bam::{SamFlag, check_flag};
+//! # let read_flag = record.flag(); in general this is the way to obtian a flag.
+//! let read_flag = 64;
+//! assert_eq!(check_flag(read_flag, SamFlag::FIRST_IN_PAIR, SamFlag::MATE_UNMAPPED), true);
+//! ```
+
+
+///
+/// This structure contains constants representing SAM flag values as u16.
+/// Using this structure incurs no runtime cost.
+/// 
+/// ```
+/// use rust_htslib::bam::{SamFlag};
+/// to get the value of a flag representing a read paired, and reversly mapped.
+/// let flag = SamFlag::PAIRED + SamFlag::READ_RERVERSE;
+/// 
+/// ```
+pub struct SamFlag;
+
+
+impl SamFlag {
+    pub const PAIRED: u16 = 1;
+    pub const PROPERLY_PAIRED: u16 = 2;
+    pub const READ_UNMAPPED: u16 = 4;
+    pub const MATE_UNMAPPED: u16 = 8;
+    pub const READ_RERVERSE: u16 = 16;
+    pub const MATE_REVERSE: u16 = 32;
+    pub const FIRST_IN_PAIR: u16 = 64;
+    pub const SECOND_IN_PAIR: u16 = 128;
+    pub const NOT_PRIMARY_ALN: u16 = 256;
+    pub const FAIL_QC: u16 = 512;
+    pub const DUPLICATE: u16 = 1024;
+    pub const SUPPLEMENTARY: u16 = 2048;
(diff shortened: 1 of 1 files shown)


johanneskoester (changes requested):
Nice!

src/bam/mod.rs
```diff
@@ -11,6 +11,7 @@ pub mod header;
 pub mod index;
 pub mod pileup;
 pub mod record;
+pub mod sam_flag;
```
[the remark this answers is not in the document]
rLannes: Thank you for your return! I changed the name of both the module and the Structure. And ran cargo fmt. (Sorry for that, it is the first time I am contributing to a rust repo). 
Romain,

johanneskoester (changes requested):
Isn't the sam_flag module just a repetition of flags now? Did you mean to remove it?
Also, would it make sense to have two separate methods `is_in()` and `is_not_in()` for the Flags struct instead of check_flags?

rLannes:
Thank you for your return.  

1- I am not sure what you are referring to with this: "Isn't the sam_flag module just a repetition of flags now? Did you mean to remove it?". I thought I did by changing the sam_flag module name to flags, as suggested. Is that not the case?

As of the latest commit to use this module, one has to import: ```use rust_htslib::bam::flags;``` Did I miss something, or fail to update/commit??

2- I often have to test for both the presence and absence of a flag, but now that you mentioned it, it is clearly confusing for an API. I will split this function into two and change the doc string accordingly.

Thank you for your time and suggestion!

rLannes:
I added two functions (removed the old one) and updated the docstring accordingly.
 You now use them like this:
```
use rust_htslib::bam::flags{Flag, is_in_flag, is_not_in_flag};
//let read_flag = record.flag(); // In general, this is the way to obtain a flag.
let read_flag = 64;
assert_eq!(is_in_flag(read_flag, Flag::FIRST_IN_PAIR), true);
assert_eq!(is_not_in_flag(read_flag, Flag::MATE_UNMAPPED), true);
```

johanneskoester:
See the tab "files changed" at the top. There is still the old sam_flags.rs in this PR.

rLannes:
Thank you for catching that, I was not aware I had to "git rm" old files to get rid of them. It should be good now.

src/bam/flags.rs
```diff
@@ -0,0 +1,129 @@
...
+    //! assert_eq!(flags::is_not_in_flag(read_flag, Flag::NOT_PRIMARY_ALN + Flag::FAIL_QC), true);
+    //! ```
+    //! let test that the read is mapped.
+    //! ```
+    //!
+    //! use rust_htslib::bam::flags::{Flag, is_not_in_flag};
+    //! let read_flag = 18;
+    //! assert_eq!(is_not_in_flag(read_flag, Flag::READ_UNMAPPED), true);
+    //! ```
+    //!
```
johanneskoester, line 60: These comments should be put above the function, without the exclamation mark but with three slashes instead: https://doc.rust-lang.org/rust-by-example/meta/doc.html

src/bam/flags.rs
```diff
@@ -0,0 +1,129 @@
...
+    //! assert_eq!(is_in_flag(read_flag, Flag::PAIRED + Flag::FIRST_IN_PAIR), true);
+    //! ```
+
+    if (in_ & flag) != in_ {
+        return false;
+    }
+    true
+}
+
+/* Deprecated
```
johanneskoester, line 86: remove the deprecated code?

johanneskoester (changes requested):

johanneskoester:
Something is wrong now.

rLannes:
I am confused None of the linting errors have to do with the flag module.

PR #455, opened 2024-11-22
fix!: memory leak in faidx when fetching sequences

brentp:
this changes the API. I don't see a way to still return `Result<&[u8]>`


johanneskoester (approved):

johanneskoester:
Thank you!

PR #465, opened 2025-02-17
ci: Small cleanup of the Github Actions in the PR workflow

martin-g:
* replace the not maintained `actions-rs/toolchain` action with `dtolnay/rust-toolchain`
* replace `actions-rs/cargo` with direct usage of `cargo`
* * replace `actions-rs/clippy` with direct usage of `cargo clippy` (it even shows that there are some problems in the code)
* add build on Linux ARM64. Often Bioconda recipes fail to build on linux-aarch64 because they use a version of rust-htslib that has/had problems for this platform

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,9 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -10,7 +10,7 @@ jobs:
     if: github.repository_owner == 'rust-bio'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
         with:
           submodules: recursive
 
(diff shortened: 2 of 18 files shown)


brainstorm (approved):
Very much needed, thanks @martin-g ;)

martin-g:
> Please fix the lint errors if you are ok doing so?

Doing it!

martin-g:
OK! All is green!
Please review when you have time!

brainstorm:
Thanks much @martin-g, LGTM.

@johanneskoester, any objections?

brainstorm:
@martin-g Actually, can you squash your commits a bit? CI tests tend to be a bit verbose on commit history, other than that the changes look great and code looks even better now too ;)

dlaehnemann:
@martin-g We are [implementing the same changes over in the `rust-bio` repository](https://github.com/rust-bio/rust-bio/pull/626) and I was wondering if it was a deliberate choice to set the runners in `rust.yml` to `ubuntu-22.04` instead of `ubuntu-latest`? And if so, why?

martin-g:
The OS images (22.04 vs 24.04) come with different set of preinstalled software. 
Some package was missing in the ARM64 image but I don't remember which one exactly. 
Try with -latest first and use it if it works!

PR #476, opened 2025-05-19
feat: Allow for non-diploid genotypes

### LINKED ISSUE #474: dealing with mixed ploidy
As made clear in the documentation, this crate does not support anything other than diploid calls. Being able to deal with mixed ploidy would be super useful, e.g., on the mammalian X chromosome.

I have figured out how to get around this, but I sadly do not have time right now to add this functionality to the crate. I'm hoping to get to it later, but in the meantime, I wanted to write some notes here for the benefit of anyone else dealing with this issue. I hope this is helpful to someone and thanks for everyone's work on this library!

### Reading

Instead of using `record.genotypes()`, you have to use `record.format(b"GT").integer()`. The length of the returned buffered vector divided by the number of samples is the number of alleles per sample, so in the most common case of a mix of diploid and haploid calls, `format_ints[i*2:i*2+2]` will contain the encoded representations of the two alleles for the i-th sample. For diploid calls, both of these ints can be converted to `GenotypeAllele` using `GenotypeAllele::from()`. For haploid calls, the first int can be converted to a `GenotypeAllele` the same way, but the second int will be `0x80000001` to signal that this sample only has a single allele, and if you try to pass that number to `GenotypeAllele::from()` it will panic.

So, for example, `[0x02, 0x80000001, 0x02, 0x04]` would represent the two calls `0` and `0/1`.

### Writing

Instead of using `record.push_genotypes()`, you have to use `record.push_format()` directly, with half-genotypes represented by `0x80000001` in the second position. Both of these functions cause a core dump if you don't give them a vector with length evenly divisible by the number of samples. So, for example, if you are trying to write a record on the X chromosome where the first sample is a male with allele 1 and the second sample is a female with genotype 0/1, you can do:

```rust
const BCF_END_OF_VECTOR: i32 = 0x80000001u32 as i32;
record.push_format_integer(
    b"GT",
    &[
        i32::from(Unphased(1)), 
        BCF_END_OF_VECTOR,
        i32::from(Unphased(0)),
        i32::from(Unphased(1)),
    ],
);
```

jeeeesper:
Should close #474

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -358,9 +358,9 @@ impl Record {
 
         let orig_aux_offset = self.qname_capacity()
             + 4 * self.cigar_len()
-            + (self.seq_len() + 1) / 2
+            + self.seq_len().div_ceil(2)
             + self.seq_len();
-        let new_aux_offset = q_len + extranul + cigar_width + (seq.len() + 1) / 2 + qual.len();
+        let new_aux_offset = q_len + extranul + cigar_width + seq.len().div_ceil(2) + qual.len();
         assert!(orig_aux_offset <= self.inner.l_data as usize);
         let aux_len = self.inner.l_data as usize - orig_aux_offset;
         self.inner_mut().l_data = (new_aux_offset + aux_len) as i32;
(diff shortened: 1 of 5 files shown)


brainstorm:
Thanks @jeeeesper! Would you mind fixing up the clippy issues and add some test(s)?

PR #478, opened 2025-07-11
feat(bcf): Reduce allocations by using CStr8

killercup:
As of today, every time a field is written, a `CString` is allocated as the methods only take `&[u8]` slices, which have no guaranteed `\0` termination.

## Solution

This PR puts the work on the caller -- who probably either has a static list of tags or can pre-allocate and keep references to them around. It changes many of the method signatures to take a [`&CStr8`](https://docs.rs/cstr8/0.1.4/cstr8/struct.CStr8.html), which is a type that ensures the data is both UTF-8 and null-terminated.

I don't have isolated measurements, but alongside some other allocation fixes, this improves the runtime performance of the tool I'm working with by 1.41x when writing large-ish (~500MB) BCF files.

## Alternatives considered

1. Take a `CStr`

	This is a type from the std lib, so there is no external dependency being introduced. Since the tags are also used in error messages, this would introduce UTF-8 checks, albeit in the cold path.

2. Introduce a custom type.

	The cstr8 crate seems to do what we need here but is not very popular and also at version 0.1.x. Making a new custom type would put us in charge of it, with the ability to optimize further.

3. Do nothing. This costs a bunch of performance.

## Downsides of this approach and future actions

- This introduces a new dependency. We might want to make this opaque and re-export most of it.
- Documentation needs to be updated to teach about this.
- We might still want to add a conversion trait and Cow-like type for convenience.
- This PR does not update all places where this could be done, only those that showed up in my profiling.

Please let me know what you think of this. I'm using this branch in the project I'm working on and found no issues, but I also don't use all of rust-htslib. I'll keep this branch up to date.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ serde = {version = "^1", optional = true, features = ["derive"]}
 serde_bytes = {version = "0.11", optional = true}
 thiserror = {version = "^2" }
 url = "2.5"
+cstr8 = "0.1.4"
 
 [features]
 bindgen = ["hts-sys/bindgen"]
--- a/src/bcf/header.rs
+++ b/src/bcf/header.rs
@@ -40,6 +40,7 @@ use std::str;
 
 use crate::htslib;
 
+use cstr8::CStr8;
 use linear_map::LinearMap;
 
 use crate::errors::{Error, Result};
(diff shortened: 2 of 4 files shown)


killercup:
I have a version with this and more fixes in a fork. Is anyone interested in this? Would love to upstream this

dlejeune:
Hey @killercup -- I'm working through a bunch of the PRs and I think a lot of the changes in your fork would be great to include in here. Are you still interested in merging those in? I know this PR is like a year old now :joy:

killercup:
Hi! Yeah it’s been a while 😅 I have a fork that I’m using in Rastair after
some performance profiling. Take what you want :)

dlejeune:
That's awesome. Do you want to make a PR for your whole fork, or should we just move component by component across?

killercup:
I’m on vacation right now, so if you’re motivated right now please go ahead
:) i think most of the commits are self contained

Issue #480, opened 2025-08-05
Add support for bam_aux_update_str(), bam_aux_update_int(), bam_aux_update_float()


ghuls:
Add support for HTSlib `bam_aux_update_str()`, `bam_aux_update_int()`, `bam_aux_update_float()`:
https://github.com/samtools/htslib/pull/694

This allows updating an existing aux tag instead of having to do: `remove_aux` => `push_aux`

Issue #335, opened 2021-10-08
Add update_aux function to BAM records.


ghuls:
Add update_aux function to BAM records as htslib now has the following functions:

```
bam_aux_update_str
bam_aux_update_int
bam_aux_update_float
```

PR #481, opened 2025-08-05
feat: Add support for updating existing aux tags inplace.

ghuls:
Add support for updating existing aux tags inplace, by using HTSlib:
 - bam_aux_update_str
 - bam_aux_update_int
 - bam_aux_update_float
 - bam_aux_update_array functions.

Updating aux tags this way will keep the tag at the original location and data of aux tags after get automatically moved if more space was needed when updating a tag inplace.

### DIFF
--- a/src/bam/record.rs
+++ b/src/bam/record.rs
@@ -1068,6 +1068,160 @@ impl Record {
         }
     }
 
+    /// Update or add auxiliary data.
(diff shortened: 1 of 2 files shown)


ghuls:
Closes: https://github.com/rust-bio/rust-htslib/issues/480 and https://github.com/rust-bio/rust-htslib/issues/335

PR #499, opened 2025-12-19
fix: empty_header was making a deep copy of header

### LINKED ISSUE #493: empty_record clones header
`empty_record`: https://github.com/rust-bio/rust-htslib/blob/8f1cdd75c301cb1e0ae54125a33de065c2550225/src/bcf/header.rs#L516-L519

does a `bcf_hdr_dup` on the header: https://github.com/rust-bio/rust-htslib/blob/8f1cdd75c301cb1e0ae54125a33de065c2550225/src/bcf/header.rs#L521-L526

 Is this necessary? I think we can just clone the Rc instead?

I ask because this is showing up on a profiling run in a real application for me.

The empty_record is called for every `next` on a reader: https://github.com/rust-bio/rust-htslib/blob/8f1cdd75c301cb1e0ae54125a33de065c2550225/src/bcf/mod.rs#L802-L808

brentp:
closes #493

I introduced this between 0.48.0 and 0.49.0 :man_facepalming: 

this does an Rc:clone() instead of a HeaderView::clone() which was doing a deep copy of the header. this has pretty big performance consequences.

ping @johanneskoester

### DIFF
--- a/src/bcf/header.rs
+++ b/src/bcf/header.rs
@@ -513,8 +513,8 @@ impl HeaderView {
     /// Create an empty record using this header view.
     ///
     /// The record can be reused multiple times.
-    pub fn empty_record(&self) -> crate::bcf::Record {
-        crate::bcf::Record::new(Rc::new(self.clone()))
+    pub fn empty_record(self: &Rc<Self>) -> crate::bcf::Record {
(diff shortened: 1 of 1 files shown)


johanneskoester:
The clippy error is fixed in PR #501.

PR #511, opened 2026-03-30
fix: map unmapped reads (tid=-1) to correct index in slow_idxstats

nh13:
When iterating CRAM records, unmapped reads have tid == -1. Casting
this to usize wraps to usize::MAX, causing an index-out-of-bounds
panic. Map tid == -1 to the designated unmapped slot at index nref.


johanneskoester (approved):

johanneskoester:
Thank you so much!

PR #515, opened 2026-07-13
fix: add null check to tidname on HeaderView

### LINKED ISSUE #505: Segfault when using an invalid TID for a BAM HeaderView
## The bug

Hello! I found that indexing into a BAM HeaderView with a TID outside of its range of references causes a segfault. Specifically, this happens when `rust_htslib::bam::HeaderView::tid2name(…)` is called. Here's a minimal example:
```rust
fn main() {
    use ::rust_htslib::bam::{Header, HeaderView};
    let header = Header::new();
    let header_view = HeaderView::from_header(&header);
    _ = header_view.tid2name(0);
}
```
It looks to me like the problem is that `tid2name(…)` doesn't check its input before calling the C binding for htslib's `sam_hdr_tid2name`. Maybe some edit to its definition like this would help, without changing the API:
```rust
pub fn tid2name(&self, tid: u32) -> &[u8] {
    if self.target_count() > tid { 
        unsafe { ffi::CStr::from_ptr(htslib::sam_hdr_tid2name(self.inner, tid as i32)).to_bytes() }
    } else {
        panic!(<some error message>)
    }
}
```
But I'm no expert on the inner workings of htslib, so I'm not certain that that's a complete solution. Just a guess.

From what I can tell, the only place in rust-htslib that `tid2name(…)` is being used is in the `rust_htslib::bam::record::Record` implementation of `bio_types::genome::AbstractInterval::contig()`. That's where I ran into this issue: I called `.contig()` after setting the `Record`'s TID to an invalid value.

I hope this was helpful!

## Some version details

rust-htslib version: 0.51.0

output from `cargo -v version`:
```
cargo 1.92.0 (Homebrew)
release: 1.92.0
host: aarch64-apple-darwin
libgit2: 1.9.2 (sys:0.20.2 system)
libcurl: 8.7.1 (sys:0.4.83+curl-8.15.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 15.7.3 [64-bit]
```

output from `rustc -v version`:
```
rustc 1.92.0 (ded5c06cf 2025-12-08) (Homebrew)
binary: rustc
commit-hash: ded5c06cf21d2b93bffd5d884aa6e96934ee4234
commit-date: 2025-12-08
host: aarch64-apple-darwin
release: 1.92.0
LLVM version: 21.1.7
```

Please let me know if there's any other info I ought to provide. Thank you!

minorllama:
This fixes: https://github.com/rust-bio/rust-htslib/issues/505 by adding a null check to `tid2name`. also, added a test for the segfaulting example. Returning empty `CStr` on missing tid (when the C ffi return NULL) should also be documented.

### DIFF
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -1439,7 +1439,13 @@ impl HeaderView {
     }
 
     pub fn tid2name(&self, tid: u32) -> &[u8] {
-        unsafe { ffi::CStr::from_ptr(htslib::sam_hdr_tid2name(self.inner, tid as i32)).to_bytes() }
+        // unsafe { ffi::CStr::from_ptr(htslib::sam_hdr_tid2name(self.inner, tid as i32)).to_bytes() }
+        let ptr = unsafe { htslib::sam_hdr_tid2name(self.inner, tid as i32) };
+        if ptr.is_null() {
+            b"" 
+        } else {
+            unsafe { ffi::CStr::from_ptr(ptr).to_bytes() }
+        }
     }
 
     pub fn target_count(&self) -> u32 {
@@ -3181,6 +3187,13 @@ CCCCCCCCCCCCCCCCCCC"[..],
         ];
         let actual = reader.index_stats().unwrap();
         assert_eq!(expected, actual);
+    }
+    
+    #[test]
+    fn test_nonexistent_tidname() {
+        let header = Header::new();
+        let header_view = HeaderView::from_header(&header);
+        assert_eq!(b"", header_view.tid2name(0));
     }
 
     // #[test]
--- a/src/bam/mod.rs
+++ b/src/bam/mod.rs
@@ -1441,7 +1441,7 @@ impl HeaderView {
     pub fn tid2name(&self, tid: u32) -> &[u8] {
         let ptr = unsafe { htslib::sam_hdr_tid2name(self.inner, tid as i32) };
         if ptr.is_null() {
-            b"" 
+            b""
         } else {
             unsafe { ffi::CStr::from_ptr(ptr).to_bytes() }
         }
@@ -3187,7 +3187,7 @@ CCCCCCCCCCCCCCCCCCC"[..],
         let actual = reader.index_stats().unwrap();
         assert_eq!(expected, actual);
     }
-    
+
     #[test]
     fn test_nonexistent_tidname() {
         let header = Header::new();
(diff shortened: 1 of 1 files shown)


dlejeune (changes requested):
Hey @minorllama thanks so much for this PR. Two things:

1) Can you remove the commented out code please

2) More a thought, but would it be worth changing the return type of `tid2name` to be an `Option<&[u8]>` instead of returning an empty string? Not necessarily something to change now since that breaks the API, but maybe something to consider.

minorllama:
@dlejeune Removed commented out old code with [this commit](https://github.com/rust-bio/rust-htslib/pull/515/changes/2c30fbc7be9990c945386d4c25fb4e5b4cca33db) . 

Yes, I agree it really should return an option as the getting back an empty string can be ambiguous, but it will break the api and will involve some refactoring. 

To log, this is where I see [NULL being returned from the htslib](https://github.com/samtools/htslib/blob/b499e4c7d44eebe85b2f1c13c2d33266f169e8bb/header.c#L2413-L2427).

dlejeune (approved):

dlejeune:
Thanks @minorllama. Can you correct the errors raised by the Formatting and Linting workflows? You just need to run `cargo fmt` and `RUSTFLAGS="-Dwarnings" cargo clippy --all-features --all-targets -- -Dclippy::all -Dunused_imports` and then fix the issues that come up there. That should appease the CI workflow

minorllama:
@dlejeune 
```bash
$ cargo fmt ; RUSTFLAGS="-Dwarnings" cargo clippy --all-features --all-targets -- -Dclippy::all -Dunused_imports ; git add -u -v
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
add 'src/bam/mod.rs'
```
and pushed. 

I was looking for a `contrib.md` in the root project for specs as these, but could not find them. Is there a `contributing` document that can be linked/added?

dlejeune:
Thanks @minorllama. I see it's still failing on the linting but it's not related to your code so I'm happy with this this PR. 

Yeah, there's no contrib.md. It's on my TODO list to write one at some point, but if you want to submit one then please do. 

@johanneskoester I'm happy for this PR to be merged since the linting errors aren't related to the code changes.

johanneskoester (approved):

johanneskoester:
Thanks a lot!

